PR #3, opened 2024-02-12
Add Sonarcloud

joachimdekker:
This PR proposes to add Sonarcloud to the build pipeline. I have no idea how GitHub Actions work, so I just pasted the pipeline provided by Sonarcloud and added the dotnet commands. 

@QuakeEye, please review it 🙏


.github/workflows/pr-build.yml
SilasPeters, line 1: Perhaps some more explanation of why we need certain tasks?

SilasPeters (changes requested):

SilasPeters (dismissed):

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,69 @@
...
+    branches:
+      - main
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    name: Build and analyze
+    runs-on: windows-latest
+    steps:
+      - name: Set up JDK 17
```
QuakeEye, line 13: Would maybe help to specify in the name that this is installed for sonarcloud as mentioned previously

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions/setup-java@v3
+        with:
+          java-version: 17
+          distribution: 'zulu' # Alternative distribution options are available.
```
QuakeEye: I do not think we will change this, the comment can probably be removed

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,69 @@
...
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
```
QuakeEye: This is a github thing, I don't think specifying the fetch depth is even necessary at all, as we just want to retrieve everything in this pull request

Aplib.Core/Main.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace Aplib.Core
+{
+    public class SonarTest
```
SSelinn: Change class name to used format. Same as filename. :D

Aplib.Core/Main.cs
```diff
@@ -0,0 +1,21 @@
...
+        public void Main()
+        {
+            Console.WriteLine("Hello World!");
+            int i = 0;
+
+            while (i < 10)
+            {
+                Console.WriteLine(i);
+                i++;
+            }
```
SSelinn: Why did you not use a for-loop? Any reason to use a while-loop?

Aplib.Core/Main.cs
```diff
@@ -0,0 +1,21 @@
...
+
+            Console.WriteLine("i is 1");
```
SSelinn: Isn't "i" a variable? Are you checking if i == 1 or are you stating it?

Aplib.Tests/Aplib.Tests.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
SSelinn, line 4: Was this not something else or was it 8?

Aplib.Tests/SonarTest.cs
```diff
@@ -0,0 +1,13 @@
+namespace Aplib.Tests;
+
+public class SonarTest
```
SSelinn, line 3: Name format o.o

SSelinn (dismissed):
:D

Aplib.Tests/Aplib.Tests.csproj
joachimdekker: The testing project can be in .NET 8, as .NET 8 can import .NET standard 2.1 libraries and test them.

Aplib.Tests/SonarTest.cs
joachimdekker: What do you mean? Also, what do you propose?

SilasPeters (dismissed):
LGTM!

QuakeEye (changes requested):

.github/workflows/pr-build.yml
joachimdekker: Isn't that a bit redundant?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,67 @@
...
+          .\.sonar\scanner\dotnet-sonarscanner begin
+          /k:"team-zomsa_aplib.net" /o:"team-zomsa"
+          /d:sonar.token="${{ secrets.SONAR_TOKEN }}"
+          /d:sonar.host.url="https://sonarcloud.io"
+          /d:sonar.cs.vscoveragexml.reportsPaths=coverage.xml
+      - name: Build the project
+        shell: powershell
+        run: |
+          dotnet build --no-incremental
+
```
QuakeEye:
```suggestion
```

🤨

QuakeEye (changes requested):

Aplib.Tests/SonarTest.cs
QuakeEye: This was due to the class name not being the same as the filename which had been solved, marking as resolved was alright.

QuakeEye (approved):
lgtm!

SilasPeters (approved):

PR #6, opened 2024-02-27
Add first draft of Goal

SilasPeters:
All features we considered are implemented. However, the Java implementation also mentioned `float distance` as a heuristic value, and `Object proposal` as the potential solution to the goal.

The idea of heuristic values has been included, as the BDI cycle needs this.

The concept of a "proposal" clashes with both Unity and our requirement to leave out World Models, and thus this PR excludes this.

I leave this as a draft for now, as this raises some questions for our client. I do believe that this version could be considered 'final' for this sprint.

### DIFF
--- a/Aplib.Core/Goal.cs
+++ b/Aplib.Core/Goal.cs
@@ -1,0 +1,37 @@
+namespace Aplib.Core
+{
+	public class Goal
+	{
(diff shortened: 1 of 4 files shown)


SilasPeters:
Closed because of incorrect branch name

PR #7, opened 2024-02-27
feat: add Goal

SilasPeters:
All features we considered are implemented. However, the Java implementation also mentioned `float distance` as a heuristic value, and `Object proposal` as the potential solution to the goal.

The idea of heuristic values has been included, as the BDI cycle needs this.

The concept of a "proposal" clashes with both Unity and our requirement to leave out World Models, and thus this PR excludes this.


Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        // MetaData useful for debugging
+        /// <summary>
+        /// The name used to display the current goal during debugging, logging, or general overviews.
+        /// </summary>
+        public readonly string Name;
+        /// <summary>
+        /// The description used to describe the current goal during debugging, logging, or general overviews.
+        /// </summary>
+        public readonly string Description;
```
joachimdekker: Order these fields and properties according to the [Google Style Guide](https://google.github.io/styleguide/csharp-style.html):

I don't think it is necessary to group properties based on debugging, as such, the comment above `Name` can be removed.

Directly quoted from the styleguide:
Class member ordering:
  - Group class members in the following order:
      - Nested classes, enums, delegates and events.
      - Static, const and readonly fields.
      - Fields and properties.
      - Constructors and finalizers.
      - Methods.
  - Within each group, elements should be in the following order:
      - Public.
      - Internal.
      - Protected internal.
      - Protected.
      - Private.
  - Where possible, group interface implementations together.

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <param name="goalPredicate">The predicate which defines whether a goal is reached</param>
+        /// <param name="name">The name of this goal, used to quickly display this goal in several contexts.</param>
+        /// <param name="description">The description of this goal, used to explain this goal in several contexts.</param>
+        /// <seealso cref="GoalStructure"/>
+        public Goal(Tactic tactic, IGoalPredicate goalPredicate, string name, string description)
+        {
+            _tactic        = tactic;
+            _goalPredicate = goalPredicate;
+            Name           = name;
+            Description    = description;
```
joachimdekker: There is no guideline for this, but I don't like it the alignment of these variables. It makes it pretty to read now, but when expanding the class, it can become quite hard to format it (as VS does not have an alignment feature).

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// A goal effectively combines a predicate with a tactic, and aims to meet the predicate by applying the tactic.
+        /// Goals are combined in a <see cref="GoalStructure"/>, and are used to prepare tests or do the testing.
+        /// </summary>
+        /// <param name="tactic">The tactic used to approach this goal.</param>
+        /// <param name="goalPredicate">The predicate which defines whether a goal is reached</param>
+        /// <param name="name">The name of this goal, used to quickly display this goal in several contexts.</param>
+        /// <param name="description">The description of this goal, used to explain this goal in several contexts.</param>
+        /// <seealso cref="GoalStructure"/>
+        public Goal(Tactic tactic, IGoalPredicate goalPredicate, string name, string description)
```
joachimdekker: Why does the `goalPredicate` have it's own interface? I would say that `goalPredicate` is a predicate, meaning a `Func<bool>` or `Func<Heuristics>`. What is the added bonus of an extra interface?

Aplib.Tests/Aplib.Tests.csproj
```diff
@@ -26,4 +27,8 @@
     <ProjectReference Include="..\Aplib.Core\Aplib.Core.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Tools\" />
+  </ItemGroup>
+
```
joachimdekker: I don't see the folder `Tools` in the committed files. Did you forget a commit, or can these lines be removed?

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        iterations.Should().BeGreaterThan(0);
+    }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void Goal_WhenReached_ReturnsAsCompleted(bool completed)
+    {
+        // Arrange
+        float distance = completed ? 0 : 69420;
```
joachimdekker: Test smell: We don't want conditionals in tests. 

Refactor this test as a fact for no completion and a completion, eliminating the need for a theory or parameters.

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    #region HelperMethods
+
+    private static Goal DefaultGoal(Tactic? tactic = null, IGoalPredicate? predicate = null, string? name = null,
+        string? description = null) => new(
+            tactic: tactic ?? new TacticStub(() => { }),
+            goalPredicate: predicate ?? new ConstantGoalPredicate(0),
+            name: name ?? "Such a good goal name",
+            description: description ??
+                         "\"A lie is just a good story that someone ruined with the truth.\" - Barney Stinson");
```
joachimdekker: Nice function!

Helper methods, however, should be in their own files. Additionally, I would propose creating a `TestGoalBuilder` class with a Fluent Interface so that when we add stuff to goal, we don't have to alter this method, but only have to add one method to the TestGoalBuilder class (which can later inherit from the actual GoalBuilder).

A bit like:

```cs
internal class TestGoalBuilder 
{
	private Tactic _tactic = new TacticStub(() => {});
    private IGoalPredicate _predicate = new ConstantGoalPredicate(() => {});
    private string _name = "Test Goal";
    private string _description = "Legen... wait for it - DARY ~ Barney Stinson"
    
	public TestGoalBuilder() 
    {
    }
    
    ... Methods to add 
    
	public Goal Build() {
		return new(
			_tactic,
			_predicate,
			_name,
			_description
		)
	}
}
```

If you have a better idea of implementing this class, please indicate so. I think it is a bit double to do it this way (as we have to store the tactics, names, descriptions in this class and not in `Goal`) but because `Goal` hides a lot of fields, this is the only way. We could also just subclass `Goal` and make a mock/stub out of that 🤔

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        var name = "Such a good goal name";
+        var description = "\"A lie is just a good story that someone ruined with the truth.\" - Barney Stinson";
+
+        // Act
+        Goal g = new(tactic, predicate, name, description); // Does not use helper method DefaultGoal on purpose
+
+        // Assert
+        g.Should().NotBeNull();
+        g.Name.Should().Be(name);
+        g.Description.Should().Be(description);
```
joachimdekker: For these assertions, we have to agree on one way. I am using the `Assert` static class in my PR. I am not familiar with Fluid Assertions, but they look kinda nice.

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// The <see cref="Tactic"/> used to achieve this <see cref="Goal"/>, which is executed during every iteration
+        /// of the BDI cycle.
+        /// </summary>
+        /// <seealso cref="Iterate()"/>
+        private readonly Tactic _tactic;
+        /// <summary>
+        /// The <see cref="IGoalPredicate"/> used to test whether this <see cref="Goal"/> has been completed.
+        /// </summary>
+        /// <seealso cref="IsCompleted()"/>
+        private readonly IGoalPredicate _goalPredicate;
```
joachimdekker: Nitpick: missing enter here:
```suggestion
        /// <summary>
        /// The <see cref="Tactic"/> used to achieve this <see cref="Goal"/>, which is executed during every iteration
        /// of the BDI cycle.
        /// </summary>
        /// <seealso cref="Iterate()"/>
        private readonly Tactic _tactic;

        /// <summary>
        /// The <see cref="IGoalPredicate"/> used to test whether this <see cref="Goal"/> has been completed.
        /// </summary>
        /// <seealso cref="IsCompleted()"/>
        private readonly IGoalPredicate _goalPredicate;
```

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
+namespace Aplib.Core.Desire
+{
+    public class Goal
```
joachimdekker, line 11: This class is missing documentation.

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
+using Aplib.Core.Desire;
+using Aplib.Tests.Stubs.Desire;
+using FluentAssertions;
+
+namespace Aplib.Tests.Desire;
+
+public class GoalTests
+{
+    [Fact]
+    public void Goal_WhenConstructed_ContainsCorrectMetaData()
```
joachimdekker, line 16: A comment for every tests would be nice, describing the test in the style of Given, When, Then.

Example:
```suggestion
    /// <summary>
    /// Given valid parameters and metadata,
    /// When the goal is constructed,
    /// Then the goal should correctly store the metadata.
    /// </summary>
    [Fact]
    public void Goal_WhenConstructed_ContainsCorrectMetaData()
```

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        /// <summary>
+        /// The goal is considered to be completed, when the distance of the
+        /// <see cref="CurrentHeuristics"/> is below this value.
+        /// </summary>
+        private const float Epsilon = 0.005f;
+
+        /// <summary>
+        /// The <see cref="Heuristics"/> of the current state of the game.
```
joachimdekker: Documentation for properties should always begin with `gets or sets` (depending on what is available).

Example:
```suggestion
        /// Gets the <see cref="Heuristics"/> of the current state of the game.
```

Aplib.Core/Desire/Goal.cs
joachimdekker: Make these fields auto-properties. It is uncommon to have `public readonly` fields, as auto-properties are a newer feature and thus should be used.

They also enable quick alteration of the visibility of getters and setters.
Also: an enter was forgotten between the fields.

```suggestion
        // MetaData useful for debugging
        /// <summary>
        /// The name used to display the current goal during debugging, logging, or general overviews.
        /// </summary>
        public string Name { get; }

        /// <summary>
        /// The description used to describe the current goal during debugging, logging, or general overviews.
        /// </summary>
        public string Description { get; }
```

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
+namespace Aplib.Core.Desire
+{
+    public class Goal
+    {
+        /// <summary>
+        /// The goal is considered to be completed, when the distance of the
+        /// <see cref="CurrentHeuristics"/> is below this value.
+        /// </summary>
+        private const float Epsilon = 0.005f;
```
joachimdekker: Is it necessary to make this a constant value? I would argue that for different types of heuristics, one might want to change the epsilon value. 

Furthermore, as we are working with quite small values, would a `double` type be better?

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// A goal effectively combines a predicate with a tactic, and aims to meet the predicate by applying the tactic.
+        /// Goals are combined in a <see cref="GoalStructure"/>, and are used to prepare tests or do the testing.
+        /// </summary>
+        /// <param name="tactic">The tactic used to approach this goal.</param>
+        /// <param name="goalPredicate">The predicate which defines whether a goal is reached</param>
+        /// <param name="name">The name of this goal, used to quickly display this goal in several contexts.</param>
+        /// <param name="description">The description of this goal, used to explain this goal in several contexts.</param>
+        /// <seealso cref="GoalStructure"/>
```
joachimdekker: `GoalStructure` is not defined yet, so we should not yet mention it in the code.

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,79 @@
...
+        public Goal(Tactic tactic, IGoalPredicate goalPredicate, string name, string description)
+        {
+            _tactic        = tactic;
+            _goalPredicate = goalPredicate;
+            Name           = name;
+            Description    = description;
+
+            CurrentHeuristics = _goalPredicate.Test(); // TODO is this the right time?
```
joachimdekker: Probably not the right time, let the agent handle the CurrentHeuristics. We can make it so that CurrentHeuristics has a default value we can set during the constructor.

Aplib.Core/Desire/Goal.cs
SSelinn: Not sure if I missed a memo, but didn't we do camelCase for variables? So wouldn't this be epsilon? There are some more variables that start with an uppercase letter so ill just comment this one.

SilasPeters: Epsilon is a `const`, thus it starts with a capital afaik. I don't know which other variables you mean. Public variables should also start with a capital afaik

Aplib.Core/Desire/IGoalPredicate.cs
```diff
@@ -0,0 +1,15 @@
+namespace Aplib.Core.Desire
+{
+    /// <summary>
+    /// This is called a predicate, but really it represents the way to calculate the current state's heuristic values.
```
SSelinn: If the name isn't representative of what it does can't we call it something like "IHeuristicState" or "IHeuristicProgress" instead?

Aplib.Core/Desire/Tactic.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <see cref="Action"/>s, which are executed in a Believe Desire Intent Cycle.
+    /// </summary>
+    /// <seealso cref="Goal"/>
+    /// <seealso cref="Action"/>
+    public abstract class Tactic
+    {
+        /// <summary>
+        /// Execute the next cycle in the Believe Desire Intent Cycle.
+        /// </summary>
+        public abstract void IterateBdiCycle();
```
SSelinn, line 15: Maybe my personal preference, but capitalize BDI since we capitalize the start of all words, like IterateBDICycle()? Or don't we do this because it is an afkorting?

Aplib.Core/Desire/Goal.cs
joachimdekker: According to the [C# Identifier Name](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions) document, all constant values should be named using PascalCase.

> Use PascalCase for constant names, both fields and local constants.

Aplib.Core/Desire/Tactic.cs
joachimdekker: According to this [SO answer](https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms), if there is an abreviation of three letters or more, this is well-named.

> The .NET Framework guidelines recommend that acronyms three letters or longer use mixed case for PascalCase and camelCase identifiers, except for the first word of a camelCase identifier. Thus for a class name you might have XmlDocument, while a local variable might be named httpRequest

joachimdekker (changes requested):
Nice foundation! I have left a couple of comments, ranging from simple nitpicks/docs to critical questions.

Aplib.Tests/Stubs/Desire/ConstantGoalPredicate.cs
```diff
@@ -0,0 +1,13 @@
+using Aplib.Core.Desire;
+
+namespace Aplib.Tests.Stubs.Desire;
+
+/// <summary>
+/// A <see cref="IGoalPredicate"/> which always returns <see cref="Heuristics"/> with the same distance.
+/// </summary>
+/// <param name="distance">The distance which teh predicate must always return.</param>
```
SSelinn: Spelling mistake: teh -> the :)

SSelinn (changes requested):
Some minor comments, but it looks fine to me otherwise

Aplib.Core/Desire/Goal.cs
JensSteenmetz: Summary of the most important keywords (I'll put this on the Wiki).

## Access modifiers.
Access modifiers in order of priority when combining them ([Microsoft C# Docs: Code-style rule options](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options?view=vs-2017)):
```cs
public
private
protected
internal
file
static
extern
new
virtual
abstract
sealed
override
readonly
unsafe
required
volatile
async
```
**Examples:**
```cs
private static readonly bool s_foo;
protected new virtual int bar;
public abstract class Baz { ... };
```

## Identifier names.
Naming conventions for identifiers ([Microsoft C# Docs: Identifier names](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names), [Microsoft C# Docs: protected](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected)):

```cs
namespace PascalCase { ... } // Namespaces.
<TPascalCase> // Generic types. Prefix 'T'.
interface IPascalCase { ... } // Interfaces. Prefix 'I'.
class PascalCase { ... } // Classes.
public class PascalCaseAttribute: Attribute { ... } // Attributes. Suffix 'Attribute'.
... PascalCase( ... camelCase ) { ... } // Methods and method arguments.
... PascalCase { get; set; } // Properties.
const PascalCase; // Constants.
... camelCase; // Local variables.
public static s_PascalCase; // Public static fields. Prefix 's_'.
protected static s_camelCase; // Protected static fields. Prefix 's_'.
private static s_camelCase; // Private static fields. Prefix 's_'.
public PascalCase; // Public non-static fields.
protected camelCase; // Protected non-static fields.
private _camelCase; // Private non-static fields. Prefix '_'.
```
**Additional naming conventions:**
- Use meaningful and descriptive names for variables, methods, and classes.
- Prefer clarity over brevity.
- Enum types use a singular noun for nonflags, and a plural noun for flags.
- Identifiers shouldn't contain two consecutive underscore (_) characters. Those names are reserved for compiler-generated identifiers.
- Avoid using abbreviations or acronyms in names, except for widely known and accepted abbreviations.
- Use meaningful and descriptive namespaces that follow the reverse domain name notation.
- Choose assembly names that represent the primary purpose of the assembly.
- Avoid using single-letter names, except for simple loop counters.

## Order of members (we have yet to decide on this, but for example).
The ordering of class members is as follows ([StyleCop Docs](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md)).

**Within a class, struct or interface:**
- Constant Fields
- Fields
- Constructors
- Finalizers (Destructors)
- Delegates
- Events
- Enums
- Interfaces (interface implementations)
- Properties
- Indexers
- Methods
- Structs
- Classes

**Within each of these groups, order by access:**
- Public.
- Internal.
- Protected internal.
- Protected.
- Private.

**Within each of the access groups, order by static, then nonstatic:**
- Static
- Nonstatic

**Within each of the static/nonstatic groups of fields, order by readonly, then nonreadonly:**
- Readonly
- Nonreadonly

Aplib.Core/Desire/Goal.cs
JensSteenmetz: I would suggest using StyleCop guides for member ordering. See the Wiki and my other comment. We could discuss this tomorrow.

Aplib.Core/Desire/Goal.cs
JensSteenmetz: It prevents a lot of backtracking in comments later. I think you have a good point, but how would you suggest dealing with missing refs later?

Aplib.Core/Desire/Goal.cs
JensSteenmetz: I agree with Joachim on his points. Also, I have some doubts regarding the current heuristic mechanism used. I am not sure what the current goal of the heuristic mechanism is compared to a simple goal predicate. I guess that a heuristic is used to either ...:

**... allow more flexibility in goal predicates (i.e., 'termination predicates', 'when has the goal been completed?').**
- Firstly, I think the above does not define a heuristic. If the above description is the purpose of a 'heuristic' in this code, the naming is a bit confusing to me and does not align with what I would expect it to mean.
- If the above is what is meant, then the same functionality can easily be achieved by a simple boolean predicate. On the contrary, a simple boolean predicate would become unnecessarily complicated.
- A boolean predicate, or even a guard, would probably make it significantly easier for the user to reason about his logic, compared to a heuristic. A few lectures can be devoted to explain what a heuristic is to _CS students_, so I don't think it would appeal to the average Unity developer.
- Long story short, if the above is the intended purpose, I would completely remove the concept of heuristic and replace it with a boolean predicate as I do not see any benefit of a heuristic if used like this.

**... decide on how to proceed in the goal structure (i.e., 'which goal to choose?')**
- One might want to use a different 'termination predicate' than a heuristic. Right now this is not possible (or at the very best considerably difficult to achieve).
- Same point as for the other interpretation of heuristic, but for completeness I will put it here again: a boolean predicate, or even a guard, would probably make it significantly easier for the user to reason about his logic, compared to a heuristic. A few lectures can be devoted to explain what a heuristic is to _CS students_, so I don't think it would appeal to the average Unity developer.
- Another point that applies here as well: the same functionality can easily be achieved by a simple boolean predicate. On the contrary, a simple boolean predicate would become unnecessarily complicated.
- Long story short, for the above purpose, I think the idea of a heuristic is a great idea! However, I think it should be optional, maybe even only part of a separate special Goal class. I would for now remove the concept of heuristic from the goal class, remove it with a boolean predicate, and add the idea of a heuristic for goal structure decisions to the backlog to be researched (I think it has great potential for extended functionality).

Aplib.Core/Desire/IGoalPredicate.cs
JensSteenmetz: I agree, the code comment suggests there should be a name change.

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        // Arrange
+        Tactic tactic = new TacticStub(() => { });
+        IGoalPredicate predicate = new ConstantGoalPredicate(0f);
+        var name = "Such a good goal name";
+        var description = "\"A lie is just a good story that someone ruined with the truth.\" - Barney Stinson";
```
JensSteenmetz: Nitpick: I would argue `var` instead of `string` does not improve code readability. Also see [C# at Google Style Guide: the var keyword](https://google.github.io/styleguide/csharp-style.html#the-var-keyword).

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    [Fact]
+    public void Goal_WhenIterating_DoesIterate()
+    {
+        // Arrange
+        var iterations = 0;
+        Tactic tactic = new TacticStub(() => iterations++);
+
+        // Act
+        Goal g = DefaultGoal(tactic: tactic);
```
JensSteenmetz: No single-letter names as per Microsoft C# naming conventions.

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    public void Goal_WhenConstructed_ContainsCorrectMetaData()
+    {
+        // Arrange
+        Tactic tactic = new TacticStub(() => { });
+        IGoalPredicate predicate = new ConstantGoalPredicate(0f);
+        var name = "Such a good goal name";
+        var description = "\"A lie is just a good story that someone ruined with the truth.\" - Barney Stinson";
+
+        // Act
+        Goal g = new(tactic, predicate, name, description); // Does not use helper method DefaultGoal on purpose
```
JensSteenmetz: No single-letter names as per Microsoft C# naming conventions.

JensSteenmetz (changes requested):

Aplib.Core/Desire/Goal.cs
SilasPeters: @SSelinn does this resolve this conversation?

Aplib.Core/Desire/IGoalPredicate.cs
SilasPeters: I agree with your comment. The comment should mentioned, however, that I chose this name to match Aplib's old code and the paper. We thus have a choice in naming, and to prevent confusion in naming I chose the old name, and to prevent confusion in usage I added this comment. We now have a choice.

Aplib.Core/Desire/Tactic.cs
SilasPeters: It's cursed, but code warnings are more cursed. Rider doesn't like names like `CustomerID` either...

Aplib.Core/Desire/Goal.cs
SilasPeters: *Laughs pitiful in Rider*. Sure I will fix this

Aplib.Tests/Aplib.Tests.csproj
SilasPeters: Git did not include this as the folder was empty. I will remove this

Aplib.Tests/Desire/GoalTests.cs
SilasPeters: Agreed, but the editorconfig should warn me for this

Aplib.Core/Desire/IGoalPredicate.cs
JensSteenmetz: Let's discuss it in a short meeting.

Aplib.Tests/Stubs/Desire/ConstantGoalPredicate.cs
SilasPeters: Forgive me, for I have sinned...

Aplib.Tests/Desire/GoalTests.cs
JensSteenmetz: True

Aplib.Tests/Desire/GoalTests.cs
SilasPeters: See other comment

Aplib.Tests/Desire/GoalTests.cs
SilasPeters: Good call

Aplib.Core/Desire/Goal.cs
SilasPeters: My idea exactly. Its scuffed, but we know it will exist in the future. Perhaps add a TODO?

Aplib.Core/Desire/Goal.cs
SilasPeters: Abstracting implementation from the usage. I could not quite tell yet whether we just want a literal predicate or a class which handles it, with differing constructors. Perhaps not even a boolean is returned, but `Heuristics`. Chances are we will use just boolean like things which can be presented through a `Func<bool>`, but then we can change that in the future. Abstracting is more agile for now.

Aplib.Tests/Desire/GoalTests.cs
SilasPeters: That's a very neat addition!

Aplib.Core/Desire/Goal.cs
JensSteenmetz: Abstracting is agile, yes, but this is abstraction that the user will need to interact with. I think that the user should not have to deal with this.

Aplib.Core/Desire/Goal.cs
JensSteenmetz: Please look at this before resolving the conversation :).

SilasPeters: Yeah I accidentally resolved this issue, and could not find it back. Will ook into this tomorrow

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,105 @@
+namespace Aplib.Core.Desire
+{
+    /// <summary>
+    /// A goal effectively combines a heuristicFunction with a tactic, and aims to meet the heuristicFunction by
```
joachimdekker: Let's try to leave as much implementation details such as variable names out of the summary:

```suggestion
    /// A goal effectively combines a heuristic function with a tactic, and aims to meet the heuristic function by
```

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// </summary>
+        protected double epsilon;
+
+        /// <summary>
+        /// The abstract definition of what is means to test the Goal's heuristicFunction. Returns <see cref="Heuristics"/>, as
+        /// they represent how close we are to matching the heuristicFunction, and if the goal is completed.
+        /// </summary>
+        /// <seealso cref="Goal.Evaluate"/>
+        /// <remarks>The paper mentions predicates, yet the Java Aplib uses heuristics. We use heuristics as well.</remarks>
+        public delegate Heuristics HeuristicFunction();
```
joachimdekker: According to the C# Google Styleguide, nested classes, enums, delegates and events must be at the top of every class.

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,105 @@
...
+    /// the testing.
+    /// </summary>
+    /// <seealso cref="GoalStructure"/>
+    public class Goal
+    {
+        /// <summary>
+        /// The goal is considered to be completed, when the distance of the <see cref="CurrentHeuristics"/> is below
+        /// this value.
+        /// </summary>
+        protected double epsilon;
```
joachimdekker: According to the C# Google Styleguide, protected fields must come after public fields.

Aplib.Core/Desire/Goal.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// The backing field of <see cref="Heuristics"/>.
+        /// </summary>
+        private Heuristics? _currentHeuristics;
+
+        /// <summary>
+        /// The <see cref="Tactic"/> used to achieve this <see cref="Goal"/>, which is executed during every iteration
+        /// of the BDI cycle.
+        /// </summary>
+        /// <seealso cref="Iterate()"/>
+        private readonly Tactic _tactic;
```
joachimdekker, line 56: According to the Google C# Styleguides, static, const and readonly fields must come before normal fields or properties.

Aplib.Core/Desire/Goal.cs
JensSteenmetz: No problem, I should have put it in a different conversation

Aplib.Core/Desire/Goal.cs
```diff
@@ -1,59 +1,86 @@
...
+        protected double epsilon;
 
         /// <summary>
-        /// The <see cref="Heuristics"/> of the current state of the game.
+        /// The abstract definition of what is means to test the Goal's heuristicFunction. Returns <see cref="Heuristics"/>, as
+        /// they represent how close we are to matching the heuristicFunction, and if the goal is completed.
         /// </summary>
-        public Heuristics CurrentHeuristics { get; private set; }
+        /// <seealso cref="Goal.Evaluate"/>
+        /// <remarks>The paper mentions predicates, yet the Java Aplib uses heuristics. We use heuristics as well.</remarks>
```
joachimdekker: I don't think design choices should be added to xml documentation. If this is to be preserved, an addition to the Architectural documentation could be made.

Aplib.Core/Desire/Goal.cs
```diff
@@ -1,59 +1,86 @@
...
+        /// <seealso cref="Iterate()"/>
+        private readonly Tactic _tactic;
+
+        /// <summary>
+        /// Creates a new goal with specified arguments. Upon creation, the <see cref="CurrentHeuristics"/> will be
+        /// set to <see cref="Heuristics.Default"/>.
         /// </summary>
         /// <param name="tactic">The tactic used to approach this goal.</param>
-        /// <param name="goalPredicate">The predicate which defines whether a goal is reached</param>
+        /// <param name="heuristicFunction">The heuristicFunction which defines whether a goal is reached</param>
```
joachimdekker: Let's not use parameter names in the XML documentation: 
```suggestion
        /// <param name="heuristicFunction">The heuristic function which defines whether a goal is reached</param>
```

Aplib.Core/Desire/Tactic.cs
joachimdekker, line 1: Please merge main into your branch: Tactic has already been implemented.

joachimdekker (changes requested):
Just a few nitpicks. Rest looks good!

Aplib.Core/Desire/Goal.cs
SilasPeters: Agreed!

Aplib.Core/Desire/Goal.cs
SilasPeters: Ugh, forget to change the ordering afterwards

Aplib.Core/Desire/Tactic.cs
SSelinn: Then it is fine!

SSelinn (dismissed):

Aplib.Core/Desire/Goal.cs
joachimdekker: Did we use a different styleguide, or did you forget this comment?

Aplib.Core/Desire/Tactic.cs
joachimdekker: Tactic can be removed if you have merged main into your branch!

SSelinn (dismissed):
TUTA (Thumbs Up To All)

JensSteenmetz (dismissed):
nltho (nothing left to hate on)

JensSteenmetz (dismissed):
iosgc (I only see good code)

SSelinn (dismissed):
PLMFL (Presentation Looking Mighty Fine Lad)

JensSteenmetz (dismissed):
kutgw (keep up the good work)

JensSteenmetz (approved):
iaotuh (I approve of this unparallelled humor)

Aplib.Core/Desire/Tactic.cs
SilasPeters: No, because the new tactic does not implement any `IterateBdiCycle`, which means that I get an error in my Iterate, and the Iterate is used by someone elses code in another PR. To keep this PR atomic and stand-alone, me and @JensSteenmetz descided to keep this variant of Tactic. It's bad, but the scope of this PR is not to cope with another's PR.

Aplib.Core/Desire/Goal.cs
SilasPeters: I don't see what's wrong to be honest

Aplib.Core/Desire/Tactic.cs
SilasPeters: If I switch to the new tactic (outside if desire) I can't implement ItersteBdiCycle, as the new tactic has no relevant functions. If I remove IterateBdiCycle, other code errors. Aka, we have to rethink stuff. ~~But for the love of god, let this PR get merged~~

SSelinn (approved):
OLAATP (Operation Looking All According To Plan)

joachimdekker (approved):
FTPRCBM! (Finally, This Pull Request Can Be Merged)

PR #8, opened 2024-02-27
feat: add Action

joachimdekker:
This PR proposes to add actions to the codebase. The actions are really barebones, and may need to be expanded upon in the future. This PR is purely to provide a base for other stories that need to build upon the action.


Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        // Arrange
+        Aplib.Core.Action<string> action = new();
+        string? result = null;
+        action.Effect = (query) => result = query;
+
+        // Act
+        action.Execute();
+
+        // Assert
+        Assert.Null(result);
```
JensSteenmetz: This assertion does not test the described behavior.  When the Effect delegate is _not_ invoked, the assertion still holds, since `result` is initialized with the value `null`, and set to `null` by the Effect delegate as well.

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+public class ActionTests
+{
+    /// <summary>
+    /// Given an Action instance with an Effect delegate,
+    /// When Execute is called,
+    /// Then the Effect delegate should be invoked.
+    /// </summary>
+    [Fact]
+    public void Execute_ShouldInvokeEffect()
```
JensSteenmetz: Does not follow naming standards for tests. See [Microsoft Docs on Unit Testing](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests).

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// <summary>
+    /// Given an Action instance with a Query delegate and an Effect delegate,
+    /// When Guard and Execute are called,
+    /// Then the Query delegate should be invoked and the result should be stored,
+    /// And the Effect delegate should be invoked.
+    /// </summary>
+    [Fact]
+    public void Guard_ShouldInvokeQueryAndStoreResult()
```
JensSteenmetz: Does not follow naming standards for tests. See [Microsoft Docs on Unit Testing](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests).

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        // Act
+        action.Guard();
+        action.Execute();
+
+        // Assert
+        Assert.Equal(42, result);
+    }
+
+    [Fact]
+    public void Actionable_ReturnsTrue_WhenQueryReturnsNonNullValue()
```
JensSteenmetz: Does not follow naming standards for tests. See [Microsoft Docs on Unit Testing](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests).

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        // Act
+        bool result = action.Actionable;
+
+        // Assert
+        Assert.True(result);
+    }
+
+    [Fact]
+    public void Actionable_ReturnsFalse_WhenQueryReturnsFalse()
```
JensSteenmetz: Does not follow naming standards for tests. See [Microsoft Docs on Unit Testing](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests).

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        // Act
+        bool result = action.Actionable;
+
+        // Assert
+        Assert.False(result);
+    }
+
+    [Fact]
+    public void Actionable_ReturnsFalse_WhenQueryReturnsNull()
```
JensSteenmetz: Does not follow naming standards for tests. See [Microsoft Docs on Unit Testing](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests).

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// The effect of the action.
+        /// </summary>
+        public System.Action<TQuery>? Effect { private get; set; }
+
+        public bool Actionable => Query!.Invoke() is not (false or null);
+
+        /// <summary>
+        /// The query of the action. Can return a value, which is stored and can be used in the effect.
+        /// </summary>
+        /// <remarks>A boolean value can also be used. If an object is used, use null to mark the action as unqualified.</remarks>
```
JensSteenmetz: What do you mean by 'unqualified'? Do you mean 'not actionable'?

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class Action<TQuery>
+    {
+        private TQuery? _storedResult;
+
+        /// <summary>
+        /// The effect of the action.
+        /// </summary>
+        public System.Action<TQuery>? Effect { private get; set; }
+
+        public bool Actionable => Query!.Invoke() is not (false or null);
```
JensSteenmetz: When `Query` is `null`, this gives a null reference exception. Is this desired behaviour? I don't think it is (I think it should simply not be actionable, maybe give a warning to the user), but if so (probably to discourage not specifying a `Query` explicitly?), instead could consider having a constructor that requires a `Query` as an argument, with default value `null`. That way it is more explicit. I don't think it should be possible for the user to create an instance of a class that throws errors, even when it is not 'finished' yet.

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// The query of the action. Can return a value, which is stored and can be used in the effect.
+        /// </summary>
+        /// <remarks>A boolean value can also be used. If an object is used, use null to mark the action as unqualified.</remarks>
+        public Func<TQuery>? Query { private get; set; }
+
+        /// <summary>
+        /// Execute the action against the world.
+        /// </summary>
+        public void Execute() => Effect?.Invoke(_storedResult!);
```
JensSteenmetz: `_storedResult` might be `null`, so I think it doesn't really make sense to provide this to the `Effect` delegate on invoke, when really it should not execute the `Effect` at all when the result has not been queried.

Edge case to consider: when you make `TQuery` a nullable type, and assign null to `_storedResult`, that would then not execute the `Effect`, while you might actually mean that the _value_ of `_storedResult` is `null`, if that makes any sense. In other words, `_storedResult` being `null` should be different than the result of the `Query` being `null`.

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Guard the action against unwanted execution. The result is stored and can be used in the effect.
+        /// </summary>
+        public void Guard() => _storedResult = Query!.Invoke();
```
JensSteenmetz: Same comment as above, on `Actionable`.

JensSteenmetz (changes requested):

Aplib.Core/Action.cs
JensSteenmetz: Also maybe specify that the `Query` should be _a lambda expression returning `null`_ to mark the action as not actionable. Or better would be, in my opinion, to allow `Query` to be null, but simply skip execution and mark the action as not actionable (as commented on `Actionable`)

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        // Arrange
+        Aplib.Core.Action<object> action = new()
+        {
+            Query = () => null
```
JensSteenmetz:
```suggestion
            Query = () => null!
```
To get rid of warning.

Aplib.Core/Action.cs
joachimdekker: Thank you for your comment. I will change the description.

As for your reply, I think it is already quite clear that it should be a _lambda expression_ returning `null` given the type of `Query`. If we allow Query to be null, a never-executed action would be created. I don't think that would be intended behaviour.

Aplib.Core/Action.cs
joachimdekker: Replying to several items in your comment:

* Query should never be null. If we use C# 10 or higher, I would mark it as `required`. Since we use C# 9, I cannot do this and this will have to do. 
* I think no warning should be given when the Query is `null`, but a exception should be thrown. I could change the exception from `NullException` to `ArgumentException` or `InvalidOperationException`. Would that solve your issue?
* We could have a constructor, but I believe it was Silas' Idea to have empty constructors so we could use the property initializer `{}` notation. We would need to discuss this.
* I think it should indeed be a compile-time error when the class is not finished yet. Sadly, we are working with an ancient version of C# (read: Unity hates developers) so we will have to make some sacrifices in order to keep the code readable (I don't think the Java Aplib is very readable).

Aplib.Core/Action.cs
joachimdekker: The null value of `_storedResult` is a placeholder that is only used in the beginning. Furthermore, using the deliberation cycles, it should be impossible to run an effect before querying the world. Hence, I don't think it is necessary to change anything here.

As for the edgecase, like I said, the `null` is a placeholder for the beginning. If the type is nullable, then there is nothing wrong with `_storedResult` being null in the beginning as the query will override the `_storedResult` when queried.

Aplib.Core/Action.cs
joachimdekker: Let's discuss this in-person or over discord another time.

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    /// <typeparam name="TQuery">The type of the query of the action</typeparam>
+    public class Action<TQuery>
+    {
+        private TQuery? _storedResult;
+
+        /// <summary>
+        /// The effect of the action.
+        /// </summary>
+        public System.Action<TQuery>? Effect { private get; set; }
```
SilasPeters, line 16: `System.` Can be removed

Tboefijn (dismissed):

Tboefijn (dismissed):

Aplib.Core/Action.cs
JensSteenmetz: - Why should Query never be `null`? I made a mistake in my previous comment; I said the Query being `null` should cause the Action to be unactionable, but I think it should be the opposite. Query should be used to guard the action. If the user doesn't want the action to be guarded, it should be possible to simply not specify a Query (then the effect would always be applied). What would be the benefit of not allowing Query to be `null`?
- I agree that giving a warning would not be ideal. Errors, however, have the same flaw in my opinion. The constructor **does** solve my issue, though, since this makes it impossible to implement such unwanted states.
- I think property initializers are well-suited for **optional** properties. Instead, constructors have the exact purpose we want here.
- Even if it was a compile-time error, it might cause confusion amongst developers, as it is not how C# is designed.

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Gets a value indicating whether the action is actionable.
+        /// </summary>
+        public bool Actionable => Query!.Invoke() is not (false or null);
+
+        /// <summary>
+        /// Gets or sets the query of the action. Can return a value, which is stored and can be used in the effect.
+        /// </summary>
+        /// <remarks>A boolean value can also be used. If an object is used, use null to mark the action as unactionable.</remarks>
+        public Func<TQuery> Query { private get; set; }
```
JensSteenmetz: Query not having a nullable type means users HAVE to implement a query. Many actions wouldn't need a guard, however, so it seems weird a Query is required. Masybe Query has a confusing naming, and should be changed to Guard (and I think the Guard method should always implicitly be called when Execute is called, or some similar behaviour).

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Action{TQuery}"/> class.
+        /// </summary>
+        /// <param name="effect">The effect of the action.</param>
+        /// <param name="query">THe query of the action.</param>
+        public Action(System.Action<TQuery> effect, Func<TQuery> query)
```
JensSteenmetz: As described above, I think the query should be nullable, and have default value null (meaning the action will always be performed when the corresponding goal is active)

Aplib.Core/Action.cs
JensSteenmetz: I understand that `_storedResult` is not meant to be null and will be overridden when queried. However, right now, the query is not implicit, so one might call `Execute` without calling `Guard` and `_storedResult` would be `null`. This would be treated in the same way as the query being called and returning `null`. The first one **shouldn't** happen, the second one **should** be possible.

Aplib.Core/Action.cs
JensSteenmetz: See my other comment on this.

JensSteenmetz (changes requested):

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Gets or sets the query of the action. Can return a value, which is stored and can be used in the effect.
+        /// </summary>
+        /// <remarks>A boolean value can also be used. If an object is used, use null to mark the action as unactionable.</remarks>
+        public Func<TQuery> Query { private get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Action{TQuery}"/> class.
+        /// </summary>
+        /// <param name="effect">The effect of the action.</param>
+        /// <param name="query">THe query of the action.</param>
```
JensSteenmetz:
```suggestion
        /// <param name="query">The query of the action.</param>
```
Grammar police

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Initializes a new instance of the <see cref="Action{TQuery}"/> class.
+        /// </summary>
+        /// <param name="effect">The effect of the action.</param>
+        /// <param name="guard">The guard of the action.</param>
+        public Action(System.Action effect, Func<bool> guard) : this(effect) => Guard = guard;
+
+        /// <summary>
+        /// Execute the action against the world.
+        /// </summary>
+        public virtual void Execute() => Effect();
```
JensSteenmetz: Maybe make this internal? It is not supposed to be called by the user, right?

Aplib.Core/Action.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Guard the action against unwanted execution. The result is stored and can be used in the effect.
+        /// </summary>
+        public virtual bool IsActionable() => Guard();
```
JensSteenmetz: Same here; make internal?

Aplib.Core/GuardedAction.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="effect">The effect of the action.</param>
+        /// <param name="guard">The guard of the action.</param>
+        public GuardedAction(System.Action<TQuery> effect, Func<TQuery?> guard) : base()
+        {
+            Effect = effect;
+            Guard = guard;
+        }
+
+        /// <inheritdoc/>
+        public override void Execute() => Effect(StoredGuardResult!);
```
JensSteenmetz: Internal

Aplib.Core/GuardedAction.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <inheritdoc/>
+        public override bool IsActionable()
```
JensSteenmetz: Internal

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// Given a side effect action with a string guard,
+    /// When the action is executed,
+    /// Then the result should be null.
+    /// </summary>
+    [Fact]
+    public void Execute_SideEffects_ReturnsCorrectEffect()
+    {
+        // Arrange
+        string? result = "abc";
+        Action action = new(effect: () => { result = "def"!; });
```
JensSteenmetz:
```suggestion
        Action action = new(effect: () => { result = "def"; });
```

Aplib.Tests/Core/ActionTests.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// Given a guarded action with an int guard,
+    /// When the action is guarded and executed,
+    /// Then the result should be the value of the guard.
+    /// </summary>
+    [Fact]
+    public void Execute_WithGuard_ShouldInvokeQueryAndStoreResult()
+    {
+        // Arrange
+        int result = 0;
+        GuardedAction<int> action = new(guard: () => 42, effect: (guard) => { result = guard; });
```
JensSteenmetz: You could consider:
```suggestion
        GuardedAction<int> action = new(guard: () => 42, effect: (guard) => result = guard);
```

JensSteenmetz (changes requested):

Aplib.Core/Action.cs
joachimdekker: Excellent suggestion!

Aplib.Core/Action.cs
joachimdekker: Wait, this is outdated. Did you review the latest changes?

JensSteenmetz: Yeah, apparently I forgot to press review when I wrote this comment last week. The other comments are fresh :)

joachimdekker: Not anymore 😛

JensSteenmetz: Ooh nice😮

JensSteenmetz (approved):
At last... LGTM!!

Tboefijn (approved):

PR #12, opened 2024-03-05
feat: add Tactic

Tboefijn:


Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+                case TacticType.Primitive:
+                    PrimitiveTactic tactic = (PrimitiveTactic)this;
+
+                    if (tactic.IsActionable())
+                        primitiveTactics.Add(tactic);
+
+                    break;
+            }
+
+            return primitiveTactics;
```
joachimdekker: I am not a fan of this switch. Another idea is to make a subclass for every type of tactic and overriding the GetFirstEnableActions() method.

Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+        private readonly Action _action;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrimitiveTactic"/> class with the specified action.
+        /// </summary>
+        /// <param name="action">The action of the primitive tactic.</param>
+        public PrimitiveTactic(Action action) : base(TacticType.Primitive, new()) => _action = action;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrimitiveTactic"/> class with the specified action.
```
JoenvandeVorle: small nitpick: write "specified action and guard" to signify difference with other constructor

Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            if (TacticType == TacticType.Primitive)
+            {
+                PrimitiveTactic tactic = (PrimitiveTactic)this;
+
+                return tactic;
+            }
+
+            return _parent.GetNextTactic();
+        }
```
joachimdekker: This method doesn't make any sense. Why would you get the next tactic from the parent?

Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+{
+    /// <summary>
+    /// Represents the type of a tactic.
+    /// </summary>
+    public enum TacticType
+    {
+        Primitive,
+        FirstOf,
+        AnyOf,
+    }
```
joachimdekker: Enums and classes should all have their seperate file. (But please first look at my other comment about subclassing tactic)

joachimdekker:
Please note that the branch name is incorrect. It should be in the format of `{scope}/{jira-id}-{name-in-kebab-case}`. In this case, I would translate it to: `feat/0007-add-tactic`

Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+            TacticType = tacticType;
+            _subTactics = new();
+            Guard = guard;
+
+            foreach (Tactic tactic in subTactics)
+            {
+                tactic._parent = this;
+                _ = _subTactics.AddLast(tactic);
+            }
+        }
```
joachimdekker: Remove the duplicate code here by using a `this` constructor:

```suggestion
        public Tactic(TacticType tacticType, List<Tactic> subTactics, Func<bool> guard)
        	: this(tacticType, subTactics)
        {
            Guard = guard;
        }
```

JoenvandeVorle (changes requested):

Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// <summary>
+    /// Represents a primitive tactic in the Aplib.Core namespace.
+    /// </summary>
+    public class PrimitiveTactic : Tactic
+    {
+        /// <summary>
+        /// Gets or sets the action of the primitive tactic.
+        /// </summary>
+        private readonly Action _action;
```
joachimdekker: How would you execute this action?

Aplib.Core/Tactic.cs
```diff
@@ -0,0 +1,167 @@
...
+        public PrimitiveTactic(Action action) : base(TacticType.Primitive, new()) => _action = action;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrimitiveTactic"/> class with the specified action.
+        /// </summary>
+        /// <param name="action">The action of the primitive tactic.</param>
+        /// <param name="guard">The guard of the tactic.</param>
+        public PrimitiveTactic(Action action, Func<bool> guard) : base(TacticType.Primitive, new(), guard) => _action = action;
+
+        public override bool IsActionable() => Guard() && _action.IsActionable();
```
joachimdekker: If the Guard is meant to mimic the base method, consider using `base.IsActionable()` instead:
```suggestion
        public override bool IsActionable() => base.IsActionable() && _action.IsActionable();
```

Aplib.Core/Tactic.cs
joachimdekker: And what is the use of a 'next' tactic?

joachimdekker (changes requested):
Good code. I have a few suggestions appended.

Aplib.Core/Tactic.cs
Tboefijn: Maybe next time write this in a suggestion?😄

Aplib.Core/Tactics/Tactic.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    /// <summary>
+    /// Represents a tactic in the Aplib.Core namespace.
+    /// </summary>
+    public abstract class Tactic
+    {
+        /// <summary>
+        /// Gets or sets the sub-tactics of the tactic.
+        /// </summary>
+        protected LinkedList<Tactic> SubTactics { get; set; }
```
joachimdekker: A primitive tactic does not have a subtactic, so it may not be best to have a list of subtactics here. Remove it here and add it in AnyOf and FirstOf, or make another subclass describing the behaviour of tactics with children.

Aplib.Core/Tactics/Tactic.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Tactic"/> class with the specified tactic type, sub-tactics, and guard.
+        /// </summary>
+        /// <param name="tacticType">The type of the tactic.</param>
+        /// <param name="subTactics">The sub-tactics of the tactic.</param>
+        /// <param name="guard">The guard of the tactic.</param>
+        protected Tactic(List<Tactic> subTactics, Func<bool> guard)
+            : this(subTactics) => Guard = guard;
```
joachimdekker: Primitve Tactic does not have subtactics, so this constructor is a bit weird. See my other comment for a possible solution.

Aplib.Core/Tactics/Tactic.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Aplib.Core.Tactics
```
joachimdekker, line 4: Kudos that you added it to a seperate namespace!

Aplib.Tests/Core/TacticTests.cs
```diff
@@ -1,63 +1,59 @@
-﻿using Aplib.Core;
+﻿using Aplib.Core.Tactics;
 using Action = Aplib.Core.Action;
-using Tactic = Aplib.Core.Tactic;
+using Tactic = Aplib.Core.Tactics.Tactic;
```
joachimdekker: This line is unnecessary. The line above is to make sure we are using our `Action` in the code, instead of `System.Action`.

Aplib.Tests/Core/TacticTests.cs
```diff
@@ -1,63 +1,59 @@
...
+    /// </summary>
     private readonly Action _emptyAction = new(() => { });
-    private static bool TrueGuard() => true;
-    private static bool FalseGuard() => false;
 
     /// <summary>
-    /// Given a tactic without a parent,
-    /// When getting the next tactic,
-    /// Then the result should be null.
+    /// Returns true.
```
joachimdekker: This comment is not very descriptive (furthermore, does it really need a docstring because it is a private method.

Aplib.Tests/Core/TacticTests.cs
joachimdekker, line 1: Could you move these tests into the Tactics subfolder? Let's adhere to the namespaces we use in the library under test (in this case `Aplib.Core.Tactics`, so the folder structure for the tests is `Aplib.Tests/Core/Tactics/*Tests.cs`)

joachimdekker (changes requested):

Aplib.Core/Tactics/FirstOfTactic.cs
```diff
@@ -8,22 +8,31 @@ namespace Aplib.Core.Tactics
     /// </summary>
     public class FirstOfTactic : Tactic
```
joachimdekker: Is FirstOfTactic not a more specific AnyOfTactic. Would it be better to subclass from that? Then you don't have to define Subtactics twice.

Aplib.Core/Tactics/Tactic.cs
```diff
@@ -8,39 +8,23 @@ namespace Aplib.Core.Tactics
...
-
-            foreach (Tactic tactic in subTactics)
-            {
-                _ = SubTactics.AddLast(tactic);
-            }
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="Tactic"/> class with the specified tactic type, sub-tactics, and guard.
+        /// Initializes a new instance of the <see cref="Tactic"/> class with the specified guard.
```
joachimdekker: the guard is not yet specified when people read this, make it 'a'
```suggestion
        /// Initializes a new instance of the <see cref="Tactic"/> class with a specified guard.
```

joachimdekker (changes requested):

joachimdekker (approved):
LGTM!

JoenvandeVorle (approved):
lgtm!

Aplib.Core/Tactics/AnyOfTactic.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Gets or sets the sub-tactics of the tactic.
+        /// </summary>
+        protected LinkedList<Tactic> SubTactics { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AnyOfTactic"/> class with the specified sub-tactics.
+        /// </summary>
+        /// <param name="subTactics">The list of sub-tactics.</param>
+        public AnyOfTactic(List<Tactic> subTactics)
```
JensSteenmetz, line 20: What is the added value of using a List as opposed to an array? Arrays also allow params.
```suggestion
        public AnyOfTactic(params Tactic[] subTactics)
```

Aplib.Core/Tactics/AnyOfTactic.cs
```diff
@@ -0,0 +1,50 @@
...
+                _ = SubTactics.AddLast(tactic);
+            }
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AnyOfTactic"/> class with the specified sub-tactics and guard condition.
+        /// </summary>
+        /// <param name="subTactics">The list of sub-tactics.</param>
+        /// <param name="guard">The guard condition.</param>
+        public AnyOfTactic(List<Tactic> subTactics, Func<bool> guard) : this(subTactics) => Guard = guard;
```
JensSteenmetz, line 35: Same as above and I feel like logically subTactics should probably be specified after the guard (which is inherently part of _this_ tactic, if that makes sense).
```suggestion
        public AnyOfTactic(Func<bool> guard, params Tactic[] subTactics) : this(subTactics) => Guard = guard;
```

Aplib.Core/Tactics/AnyOfTactic.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <inheritdoc/>
+        public override List<PrimitiveTactic> GetFirstEnabledActions()
```
JensSteenmetz, line 38: The method name does not match its functionality.

JensSteenmetz, line 38: I think this is done weirdly in Java aplib, and we should handle this differently. Semantically the `AnyOfTactic` could be renamed to `RandomTactic`, where a random `Tactic` is picked from a list of `Tactic` instances. However, this is currently not the implementation of this class.

In Java aplib, they retrieve all enabled actions (which are not even actions in reality) and choose an `Action` at random within the `Agent` class. Why should the `Agent` class handle this? Mind you, this is ONLY the case for the `AnyOfTactic`, since the other `Tactic` subclasses only yield a single `PrimitiveTactic` with this method.

Therefore, I propose to...
1. Pick an `Action` at random within this method. This is functionality exclusive to the `AnyOfTactic` class.
2. Immediately invoke the correct `Action`. After addressing point 1, this method would retrieve a single `PrimitiveTactic`. The only place this method is used is the moment that the `Action` corresponding with that `PrimitiveTactic` is invoked. So why not do that here? This is also in line with the semantics of a `Tactic`. It should simply worry about what subtactics it consists of and how to invoke them correctly (which is practically the only thing the three subclasses differ at). To elaborate on this, what I mean is a method similar to the `Execute` method of the `Action` class method (but with the correct logic of choosing an action, of course).

Aplib.Core/Tactics/FirstOfTactic.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// Initializes a new instance of the <see cref="FirstOfTactic"/> class with the specified sub-tactics and guard condition.
+        /// </summary>
+        /// <param name="subTactics">The list of sub-tactics.</param>
+        /// <param name="guard">The guard condition.</param>
+        public FirstOfTactic(List<Tactic> subTactics, Func<bool> guard) : base(subTactics, guard)
+        {
+        }
+
+        /// <inheritdoc/>
+        public override List<PrimitiveTactic> GetFirstEnabledActions()
```
JensSteenmetz, line 29: Same comments as for `AnyOfTactic`.

Aplib.Core/Tactics/PrimitiveTactic.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PrimitiveTactic"/> class with the specified action and guard.
+        /// </summary>
+        /// <param name="action">The action of the primitive tactic.</param>
+        /// <param name="guard">The guard of the tactic.</param>
+        public PrimitiveTactic(Action action, Func<bool> guard) : base(guard) => Action = action;
+
+        /// <inheritdoc/>
+        public override List<PrimitiveTactic> GetFirstEnabledActions() => IsActionable() ? new() { this } : new();
```
JensSteenmetz, line 30: Same comments as for `AnyOfTactic`.

JensSteenmetz, line 30: The `new` keyword should only be used when it is clear form the declaration what its type is. This is not the case here.

PR #13, opened 2024-03-06
feat: metadata

Laura-Berkers:
This PR takes the fields in `Goal` that are currently storing metadata about the goal out of the `Goal` class and gives them a dedicated data structure.


SilasPeters:
@LuukBerkers can you merge main?

PR #17, opened 2024-03-14
chore: add config file

QuakeEye:
This pull request adds a missing config file.. again... whoops


JoenvandeVorle (approved):
iwbniycsftasbilgtm (it would be nice if you could stop forgetting to add stuff but it looks good to me)

SSelinn (approved):
Wow (Wonderbaarlijk onderhouden woorden)

PR #18, opened 2024-03-17
feat: update Tactic structure

Tboefijn:


Aplib.Core/Tactics/FirstOfTactic.cs
```diff
@@ -21,19 +21,19 @@ public FirstOfTactic(List<Tactic> subTactics) : base(subTactics)
...
+        public override Action? GetAction()
         {
             foreach (Tactic subTactic in SubTactics)
             {
-                List<PrimitiveTactic> firstOfTactics = subTactic.GetFirstEnabledActions();
+                Action? action = subTactic.GetAction();
 
-                if (firstOfTactics.Count > 0)
-                    return firstOfTactics;
+                if (action is not null && action.IsActionable())
```
JensSteenmetz: If the Action from the subTactic was not actionable, it is null, right? So it can just be:
```suggestion
                if (action is not null)
```

Aplib.Core/Tactics/FirstOfTactic.cs
```diff
@@ -21,19 +21,19 @@ public FirstOfTactic(List<Tactic> subTactics) : base(subTactics)
...
+                    return action;
             }
 
             return new();
```
JensSteenmetz:
```suggestion
            return null;
```

Aplib.Core/Tactics/AnyOfTactic.cs
```diff
@@ -32,19 +32,22 @@ public AnyOfTactic(List<Tactic> subTactics)
...
             {
-                primitiveTactics.AddRange(subTactic.GetFirstEnabledActions());
+                Action? action = subTactic.GetAction();
+
+                if (action is not null)
+                    actions.Add(action);
             }
 
-            return primitiveTactics;
+            return actions[ThreadSafeRandom.Next(actions.Count)];
```
JensSteenmetz, line 53: If actions.Count==0, this will cause problems. This method should return `null` in that case.

Aplib.Core/ThreadSafeRandom.cs
JensSteenmetz, line 1: We don't need this class, see my other comment.

Aplib.Core/Tactics/AnyOfTactic.cs
JensSteenmetz, line 53: The System namespace already has a thread-safe static Random instance:
```suggestion
            return actions[Random.Shared.Next(actions.Count)];
```

JensSteenmetz (changes requested):
Nice changes! I have some small remarks.

Aplib.Core/Tactics/AnyOfTactic.cs
Tboefijn: I thought Random.Shared comes in .NET 6 but I can check tomorrow.

joachimdekker: Indeed, `Random.Shared` is available from .NET 6. https://learn.microsoft.com/en-us/dotnet/api/system.random.shared?view=net-8.0

Aplib.Core/ThreadSafeRandom.cs
joachimdekker: We do. Especially if we want to reduce flakeyness. We may want to implement an interface that describes a RNG.

joachimdekker (dismissed):
LGTM!

Aplib.Core/Tactics/AnyOfTactic.cs
JensSteenmetz: You are right

Aplib.Core/ThreadSafeRandom.cs
JensSteenmetz: I would not abstract RNG, that seems unnecessary. Or premature at best. But yeah, as we cannot use Random.Shared this class will do.

joachimdekker (dismissed):

JensSteenmetz (approved):
LGTM!

joachimdekker (approved):

PR #21, opened 2024-03-18
feat: add Belief and BeliefSet

JensSteenmetz:
This PR proposes to add the basic functionality of the Belief and BeliefSet classes. A Belief/BeliefSet history is _not_ stored yet. This PR aims to help get the aplib foundation working.


Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="Belief{TReference, TResource}"/> class represents a belief.
+    /// Some <i>object reference</i> is used to generate/update a <i>resource</i> 
+    /// (i.e., some piece of information on the game state as perceived by an agent).
+    /// It implements the <see cref="IBelief"/> interface.
+    /// It supports implicit conversion to <typeparamref name="TResource"/>.
+    /// </summary>
```
joachimdekker: These lines are perfect for a `remark`
```suggestion
    /// </summary>
    /// <remark>
    /// It implements the <see cref="IBelief"/> interface.
    /// It supports implicit conversion to <typeparamref name="TResource"/>.
    /// </remark>
```

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// to be used where a <typeparamref name="TResource"/> is expected.
+        /// </summary>
+        /// <param name="belief">The <see cref="Belief{TReference, TResource}"/> object to convert.</param>
+        public static implicit operator TResource(Belief<TReference, TResource> belief) => belief._resource;
+
+        /// <summary>
+        /// Generates/updates the resource if the updateIf condition is satisfied.
+        /// The resource is then updated by calling the getResourceFromReference function.
+        /// </summary>
+        public void UpdateBelief()
```
joachimdekker, line 76: Do we want the developer to update beliefs or override the `UpdateBelief` function in a subclass? If not, we might want to make this `internal`.

Aplib.Core/Belief/Beliefset.cs
```diff
@@ -0,0 +1,36 @@
...
+                GetType().GetFields()
+                .Where(field => typeof(IBelief).IsAssignableFrom(field.FieldType))
+                .Select(field => (IBelief)field.GetValue(this))
+                .ToArray();
+        }
+
+        /// <summary>
+        /// Updates all objects of type <see cref="IBelief"/> that are defined as <i>public fields</i> in <see cref="Beliefset"/>.
+        /// </summary>
+        public void UpdateBeliefs()
```
joachimdekker: Same as before. If we don't want the developer to update the beliefs, we may want to make this method internal.

joachimdekker (dismissed):
Looks good to me. Left some few remarks, but overall good code!

Aplib.Core/Belief/Belief.cs
JensSteenmetz: Good suggestion!

Aplib.Core/Belief/Belief.cs
JensSteenmetz: You are probably right! We can later reconsider this if the developer might need to access it when implementing a custom Beliefset base class for example. But we don't know this yet, so I'll make it internal.

JensSteenmetz: It's not possible because of the IBelief interface.

Aplib.Core/Belief/Beliefset.cs
JensSteenmetz: Unlike `UpdateBelief`, this method _could_ be made internal, but the user could need it when implementing their own version of `Agent` inheriting from `IAgent`. I think that we need to reconsider these accessibility levels sometime later.

QuakeEye (changes requested):
Build fails due to non-recursive rename, check the logs, there is 1 location that needs fixing

Aplib.Core/Belief/Belief.cs
joachimdekker: That's a pity.

Aplib.Core/Belief/Beliefset.cs
joachimdekker: Great, then it's settled.

joachimdekker (dismissed):
LGTM, aside from one build error, but that is a tiny mistake.

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// Some <i>object reference</i> is used to generate/update a <i>resource</i> 
+    /// (i.e., some piece of information on the game state as perceived by an agent).
+    /// </summary>
+    /// <remarks>
+    /// It implements the <see cref="IBelief"/> interface.
+    /// It supports implicit conversion to <typeparamref name="TResource"/>.
+    /// </remarks>
+    /// <typeparam name="TReference">The type of the reference used to generate/update the resource.</typeparam>
+    /// <typeparam name="TResource">The type of the resource the belief represents.</typeparam>
+    public class Belief<TReference, TResource> : IBelief
```
SilasPeters: I am missing explanation on what this 'reference' is. What does it represent? A reference to the resource, what does that look like?

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        /// <summary>
+        /// The reference used to generate/update the resource.
+        /// </summary>
+        private readonly TReference _reference;
+
+        /// <summary>
+        /// A function that takes a reference and generates/updates a resource.
+        /// </summary>
+        private readonly Func<TReference, TResource> _getResourceFromReference;
```
SilasPeters: If it generates or updates a resource, then why is it called `get...`?

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// A function that sets a condition on when the resource should be updated.
+        /// </summary>
+        private readonly Func<bool> _updateIf = () => true;
```
SilasPeters: A little leftover from initial prototyping? Seems hardcoded

Edit: it now that I see that the constructor might override this, I understand that this is a 'default' value. This can best be represented by combining both constructors, and making the `updateIf` parameter optional.
Besides, in general, it is best to express that a field is dependend on _a_ constructor, by setting it in every constructor. You move the field to where the logic around it happens.

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
+
+namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="Belief{TReference, TResource}"/> class represents a belief.
+    /// Some <i>object reference</i> is used to generate/update a <i>resource</i>
```
SilasPeters: Is generating the right word? It observes and generates a reference or memory, not the resource itself.

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// It supports implicit conversion to <typeparamref name="TResource"/>.
+    /// </remarks>
+    /// <typeparam name="TReference">The type of the reference used to generate/update the resource.</typeparam>
+    /// <typeparam name="TResource">The type of the resource the belief represents.</typeparam>
+    public class Belief<TReference, TResource> : IBelief
+    {
+        /// <summary>
+        /// The reference used to generate/update the resource.
+        /// </summary>
+        private readonly TReference _reference;
```
SilasPeters, line 21: Rename both type and field to 'resourceReference'?

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// A function that takes a reference and generates/updates a resource.
+        /// </summary>
+        private readonly Func<TReference, TResource> _getResourceFromReference;
+
+        /// <summary>
+        /// A function that sets a condition on when the resource should be updated.
```
SilasPeters: Does it set a condition, or test a condition? Same thing for in the second constructor.

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        private readonly Func<bool> _updateIf = () => true;
+
+        /// <summary>
+        /// The resource represented by the belief (i.e., some piece of information on the game state as perceived by an agent).
+        /// </summary>
+        private TResource _resource;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Belief{TReference, TResource}"/> class with a reference,
```
SilasPeters: A reference to what?

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <param name="getResourceFromReference">A function that takes a reference and generates/updates a resource.</param>
+        public Belief(TReference reference, Func<TReference, TResource> getResourceFromReference)
+        {
+            _reference = reference;
+            _getResourceFromReference = getResourceFromReference;
+            _resource = _getResourceFromReference(_reference);
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Belief{TReference, TResource}"/>
```
SilasPeters: Compared to the constructor above, this implies that it creates a smaller constructor with less contents. However, it stores even more variables.

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
+
+namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="Belief{TReference, TResource}"/> class represents a belief.
+    /// Some <i>object reference</i> is used to generate/update a <i>resource</i> 
+    /// (i.e., some piece of information on the game state as perceived by an agent).
+    /// </summary>
```
SilasPeters: I would highlight that this class represents the belief of a single object, and does not represent the general belief of the agent. Make this clearer. Specifically, you should use the word 'wrapper', as it simply represents a resource and how to update the belief of the resource.

Aplib.Core/Belief/Belief.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <param name="belief">The <see cref="Belief{TReference, TResource}"/> object to convert.</param>
+        public static implicit operator TResource(Belief<TReference, TResource> belief) => belief._resource;
+
+        /// <summary>
+        /// Generates/updates the resource if the updateIf condition is satisfied.
+        /// The resource is then updated by calling the getResourceFromReference function.
+        /// </summary>
+        public void UpdateBelief()
+        {
+            if (_updateIf()) _resource = _getResourceFromReference(_reference);
```
SilasPeters: `updateIf` reads way better as `shouldUpdate`. Perhaps rename?

SilasPeters: AFAIK we should always use the curly bracket code block

Aplib.Core/Belief/IBelief.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="IBelief"/> interface defines a contract for classes that represent a belief.
+    /// </summary>
```
SilasPeters: This does not add much information. What is your plan behind this interface, and why is it an interface and not a base-class? This is what you might explain here.

Although, why is it an interface and not a class?

Aplib.Core/Belief/Beliefset.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Linq;
+
+namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="BeliefSet"/> class can be inherited to define a set of beliefs for an agent.
+    /// </summary>
```
SilasPeters: Aside from the inheriting part, this does not add more information. For example, how should this be used? Explain that this class acts as a collection of beliefs that will be updated every frame, through reflection magic.

Aplib.Core/Belief/Beliefset.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// All <i>public fields</i> of type <see cref="IBelief"/> that are defined in the inheriting class
+    /// are automatically updated when calling <see cref="UpdateBeliefs"/>.
+    /// </remarks>
+    public abstract class BeliefSet
+    {
+        private readonly IBelief[] _beliefs;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BeliefSet"/> class.
+        /// </summary>
```
SilasPeters: I don't want to be annoying, but this again does not tell me something new.

Aplib.Core/Belief/Beliefset.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Linq;
+
+namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="BeliefSet"/> class can be inherited to define a set of beliefs for an agent.
+    /// </summary>
+    /// <remarks>
+    /// All <i>public fields</i> of type <see cref="IBelief"/> that are defined in the inheriting class
+    /// are automatically updated when calling <see cref="UpdateBeliefs"/>.
```
SilasPeters: It would make it more clear to explain that reflection is the powerhouse here. I would even move this to the summary, as this is the whole goal of the class.

Aplib.Core/Belief/Beliefset.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// All <i>public fields</i> of type <see cref="IBelief"/> that are defined in the inheriting class
+        /// are automatically updated when calling <see cref="UpdateBeliefs"/>.
+        /// </remarks>
+        protected BeliefSet()
+        {
+            _beliefs =
+                GetType().GetFields()
+                .Where(field => typeof(IBelief).IsAssignableFrom(field.FieldType))
+                .Select(field => (IBelief)field.GetValue(this))
+                .ToArray();
```
SilasPeters: Indentation is incorrect. I would append line 26 to 25.

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace Aplib.Tests.Core.Belief;
+public class BeliefTests
+{
+    private static TReference ID<TReference>(TReference reference) => reference;
+
+    private static int GetCount<T>(IEnumerable<T> reference) => reference.Count();
+
+    private static bool TrueUpdateIf() => true;
+
+    private static bool FalseUpdateIf() => false;
```
SilasPeters: Comment this. What is ID? I happen to know the haskell `id`, but that's not common knowledge.

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
+﻿using Aplib.Core.Belief;
+
+namespace Aplib.Tests.Core.Belief;
+public class BeliefTests
```
SilasPeters, line 9: Is it the norm to add a summary here? I don't think so though

Aplib.Tests/Core/Belief/BeliefTests.cs
SilasPeters: I would name these `AlwaysUpdate` and `NeverUpdate` for better visibility.

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        // Arrange
+        string reference = "def";
+        Belief<string, string> belief = new(reference, ID);
+
+        // Act
+        string resource = belief;
+
+        // Assert
+        Assert.Equal("def", resource);
```
SilasPeters: For clarity and flexibility, I would extract both string literals to a single `const`, to express that we want to test a single value to be present in both cases.

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    private static bool FalseUpdateIf() => false;
+
+
+    /// <summary>
+    /// Given a Belief instance,
+    /// When it is assigned to a variable of its resource type,
+    /// Then it is implicitly converted to its resource type.
+    /// </summary>
+    [Fact]
+    public void Belief_AssignedToResourceType_IsImplicitlyConvertedToResourceType()
```
SilasPeters: I would say `IsCorrectlyImplicitly....`

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Act
+        string resource = belief;
+
+        // Assert
+        Assert.Equal("def", resource);
+    }
+
+    /// <summary>
+    /// Given a Belief instance with an updateIf condition that is satisfied,
```
SilasPeters: If you happened to have renamed updateIf, don't forget to change it here too

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    [Fact]
+    public void UpdateBelief_UpdateIfConditionIsSatisfied_UpdatesResource()
+    {
+        // Arrange
+        List<int> list = [];
+        Belief<List<int>, int> belief = new(list, GetCount, TrueUpdateIf);
+
+        // Act
+        list.Add(69);
```
SilasPeters, line 55: Nice

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        belief.UpdateBelief();
+
+        // Assert
+        Assert.Equal(1, belief);
```
SilasPeters: Assertions.Should().Be("Fluent Assertions everywhere");

SilasPeters: Are you testing the lenght of the list here? Where does the 1 come from?

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    [Fact]
+    public void UpdateBelief_UpdateIfConditionIsNotSatisfied_DoesNotUpdateResource()
+    {
+        // Arrange
+        List<int> list = [];
+        Belief<List<int>, int> belief = new(list, GetCount, FalseUpdateIf);
+
+        // Act
+        list.Add(69);
```
SilasPeters:
```suggestion
        list.Add(420);
```

Aplib.Core/Belief/Beliefset.cs
SilasPeters, line 1: The filename has not been renamed to `BeliefSet`. Use Rider.

Aplib.Tests/Core/Belief/BeliefsetTests.cs
SilasPeters, line 1: The filename has not been renamed to `BeliefSetTests`. Use Rider.

Aplib.Tests/Core/Belief/BeliefsetTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    private class SimpleBelief : IBelief
+    {
+        public bool Updated { get; private set; } = false;
+
+        public void UpdateBelief()
+        {
+            Updated = true;
+        }
+    }
```
SilasPeters: Move this to a seperate file

Aplib.Tests/Core/Belief/BeliefsetTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        // Arrange
+        TestBeliefSet beliefSet = new();
+
+        // Act
+        beliefSet.UpdateBeliefs();
+
+        // Assert
+        Assert.True(beliefSet.Belief1.Updated);
+        Assert.True(beliefSet.Belief2.Updated);
```
SilasPeters: Explain why you expect this to be true

Aplib.Tests/Core/Belief/BeliefsetTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    private class TestBeliefSet : BeliefSet
+    {
+        public SimpleBelief Belief1 = new();
+        public SimpleBelief Belief2 = new();
+    }
```
SilasPeters: Where are the comments? How can SonarCloud approve this?

Aplib.Tests/Core/Belief/BeliefsetTests.cs
SilasPeters, line 1: You should add a test which asserts that _only_ private fields are updated, as promised by the xml comments.

Aplib.Core/Belief/Beliefset.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        private readonly IBelief[] _beliefs;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BeliefSet"/> class.
+        /// </summary>
+        /// <remarks>
+        /// All <i>public fields</i> of type <see cref="IBelief"/> that are defined in the inheriting class
+        /// are automatically updated when calling <see cref="UpdateBeliefs"/>.
+        /// </remarks>
```
JoenvandeVorle: Maybe shortly mention that this is done through reflection, for extra clarity

Aplib.Core/Belief/IBelief.cs
SilasPeters, line 1: I agree with the current implementation of a Belief. However, it misses the functionality to express inaccuracies in observation. Do we want to support this?

SilasPeters (changes requested):
Great first prototype

Aplib.Core/Belief/Belief.cs
JoenvandeVorle: agreed (that it should be renamed to shouldUpdate or similar like Silas said)

Aplib.Core/Belief/Belief.cs
JensSteenmetz: I agree, but that would be inconsistent with other classes. I proposed a chore for in the backlog to fix this.

Aplib.Core/Belief/Belief.cs
JensSteenmetz: We discussed this irl

JoenvandeVorle (changes requested):

Aplib.Core/Belief/Beliefset.cs
JensSteenmetz: I feel like the remark beneath the summary is sufficient for this. The user wouldn't (shouldn't) care about the fact that we use reflection. I think explaining this delves too deep into implementation details, and this is not the purpose of XML summaries.

Aplib.Tests/Core/Belief/BeliefTests.cs
JensSteenmetz: I added it anyway

Aplib.Core/Belief/Beliefset.cs
JensSteenmetz: Again, the use of Reflection is an implementation detail that doesn't matter when using this class.

Aplib.Tests/Core/Belief/BeliefTests.cs
JensSteenmetz: I replaced "def" by the reference. Does this solve your issue?

Aplib.Tests/Core/Belief/BeliefTests.cs
JensSteenmetz: I'm okay with this but we should discuss this in a meeting with the rest sometime.

Aplib.Tests/Core/Belief/BeliefTests.cs
JensSteenmetz: Great suggestion!

Aplib.Core/Belief/Beliefset.cs
JensSteenmetz: The use of Reflection is an implementation detail. I don't think we should mention it, as it does not add any information about functionality/usage.

Aplib.Core/Belief/Belief.cs
SilasPeters: I still believe this can best be called `TResourceReference _resourceReference`. This increases understanding without contextual knowledge.

Aplib.Tests/Core/Belief/BeliefTests.cs
SilasPeters: Altough strings are reference types, they are immutable. I would still use a dedicated `const def = "def"` to make sure no hacky code changes any pointers, but that's me who is afraid of hacky code.

Aplib.Tests/Core/Belief/BeliefTests.cs
```diff
@@ -1,87 +1,95 @@
...
-        Belief<List<int>, int> belief = new(list, GetCount, TrueUpdateIf);
+        Belief<List<int>, int> belief = new(list, reference => reference.Count, AlwaysUpdate);
 
         // Act
         list.Add(69);
         belief.UpdateBelief();
 
         // Assert
-        Assert.Equal(1, belief);
+        Assert.Equal(list.Count, belief);
```
SilasPeters: Lists are reference types, and `list` is passed down as an argument. This means that the contents of the list cannot be assumed to be unaltered. As such, if this test aims to assert that the list should contain exactly one item, comparing the belief's to the current list's count can give wrong results.

I would store the lists' current count in a variable, before creating the belief, and use that to compare. Or, just keep the `1` but explain why you use `1`.

Aplib.Core/Belief/Beliefset.cs
SilasPeters, line 1: This and other files still do not have the correct capitalization.

Aplib.Core/Belief/Beliefset.cs
SilasPeters: I agree with that implementation details must be left out. As long as the user understands that **literally** all public _Beliefs_ fields will get updated, it's fine.

Aplib.Tests/Core/Belief/BeliefsetTests.cs
```diff
@@ -1,40 +1,143 @@
...
+
+        // Assert
+        Assert.False(beliefSet.Belief1.Updated);
+        Assert.False(beliefSet.Belief2.Updated);
+    }
+
+    /// <summary>
+    /// Given a BeliefSet instance with multiple <i>private field</i> beliefs,
+    /// When UpdateBeliefs is called,
+    /// Then all beliefs are updated.
```
SilasPeters: It should say `Then no fields are updated.`

Aplib.Tests/Core/Belief/BeliefsetTests.cs
SilasPeters, line 1: The tests presented here are only homogeneous - there is no test case where a class has both public and private `IBelief`s.

Aplib.Tests/Core/Belief/BeliefsetTests.cs
```diff
@@ -1,40 +1,143 @@
...
     }
 
-    private class TestBeliefSet : BeliefSet
+    /// <summary>
+    /// Given a BeliefSet instance with multiple <i>public property</i> beliefs,
+    /// When UpdateBeliefs is called,
+    /// Then all beliefs are updated.
+    /// </summary>
+    [Fact]
+    public void UpdateBeliefs_PublicBeliefProperties_DoesNotUpdateAnyBeliefs()
```
SilasPeters: I agree that with our current definition, this test is needed. However, why are we not updating properties, and only fields?

SilasPeters (changes requested):
Also note the previous conversations which have not been resolved yet.

Aplib.Core/Belief/Beliefset.cs
JensSteenmetz: I'm not sure how to fix this. Locally they **do** have correct capitalization, but my Git says I don't have changes. Do you have any ideas?

Aplib.Tests/Core/Belief/BeliefTests.cs
JensSteenmetz: The test does not aim to assert that the list should contain exactly one item. It aims to assert that the belief is modified correctly if the shouldUpdate condition is satisfied. It does exactly that now, no matter what happens to the list.

Aplib.Core/Belief/Belief.cs
JensSteenmetz: We discussed this IRL and concluded that this would be most clear. It's not a reference to the observation per se. It is _any_ object reference that can be used to update the observation. Consider, for example, an enemy position as an observation; the reference would not be an object reference to the _enemy position_, as this is a value type and not a reference type. The reference would be the _enemy transform_. `TObservationReference _observationReference` would thus be a misleading name. I think `TReference _reference` is still the best name.

Aplib.Core/Belief/Beliefset.cs
joachimdekker: Git and capitalization changes is a big joke. Try changing the name in the Github client or on the site!

Aplib.Tests/Core/Belief/BeliefsetTests.cs
JensSteenmetz: Good eye!

Aplib.Tests/Core/Belief/BeliefsetTests.cs
JensSteenmetz: That is true, but there is no reason why the hybrid case wouldn't work. I agree that it is theoretically a possibility, but I would give tens of things more importance to be subject to testing than this, and you have to draw a line somewhere. There will always be cases you do not or cannot account for, so the important thing is that you catch the most prevalent errors.

Aplib.Tests/Core/Belief/BeliefsetTests.cs
JensSteenmetz: Allowing properties to be updated could lead to unexpected behavior very quickly. Consider a belief set as follows:
```cs
private class TestBeliefSet : BeliefSet
{
    private SimpleBelief _belief1 = new();
    
    private SimpleBelief _belief2 = new();
    
    public bool MyBool { get; set; } = true;

    public SimpleBelief MyBelief => MyBool ? _belief1 : _belief2;
}
```
On creation, MyBool is true, so the _private_ _belief1 field is added to the update array and updated on every UpdateBelief() call. Now suppose MyBool is set to false at some point. Then from that point onwards, MyBelief returns _belief2. However, _belief2 is **not** being updated as one might expect. 

There are still some catches with public fields, as they could be assigned to by the user at runtime, but it greatly restricts the number of ways users could mess up.

Aplib.Core/Belief/Beliefset.cs
JensSteenmetz: Thanks, that worked!

Aplib.Tests/Core/Belief/BeliefsetTests.cs
JensSteenmetz: Regarding how the user could mess up with public fields; we could only update _readonly public IBelief fields_. I don't see how the user could mess up then, but it might be less user-friendly having to add the readonly keyword every time.

JoenvandeVorle (dismissed):
awimo

Aplib.Tests/Core/Belief/BeliefsetTests.cs
SilasPeters: We could filter over private setters. We should thinks this more out later, but this PR has been closed. I was about to resolve all our conversations :(

Aplib.Tests/Core/Belief/BeliefTests.cs
SilasPeters: Right now you are testing something which is by definition equal: the length of X should equal to the list in belief Y, where Y is a wrapper around X. You are not testing whether it _changed_, but whether it is equal.

SilasPeters (approved):
GliMaMoVrijMiSpeBiBiBaBoO'Do

JoenvandeVorle (approved):
oaiaw (once again it appears well)

QuakeEye (approved):

PR #22, opened 2024-03-21
feat: add Metadata

Laura-Berkers:
This PR takes the fields in `Goal` that are currently storing metadata about the goal out of the `Goal` class and gives them a dedicated data structure.

It also then uses this to add a metadata field to `Tactic` and `Action`.


Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -28,14 +28,9 @@ public virtual Heuristics CurrentHeuristics(BeliefSet beliefSet)
             => _currentHeuristics ??= _heuristicFunction.Invoke(beliefSet);
 
         /// <summary>
-        /// The name used to display the current goal during debugging, logging, or general overviews.
+        /// Data about the Goal such as a name and description, this may be useful for debugging or logging.
```
joachimdekker: XML comments for public (auto) properties should start with 'Gets' (And 'Gets or sets' if set is also an option).

Also, try to summarize the property in one sentece. If you want to add extra information, add it in a <remark> tag:

```suggestion
        /// Gets or sets the metadata of the goal.
```

You can set the details in a remark:
```
/// <remark>
/// The metadata may be useful for debugging
/// </remark>
```

(For some reason, I could not set a suggestion for that, sorry.

joachimdekker: This is also the case for Action and Tactic. Please edit those accordingly.

Aplib.Core/Intent/Actions/Action.cs
```diff
@@ -21,24 +26,42 @@ public class Action
         /// Initializes a new empty instance of the <see cref="Action{TQuery}"/> class.
         /// </summary>
         /// <remarks>Only meant for internal use</remarks>
-        protected internal Action()
+        /// <param name="name">The name of this Action, used to quickly display this goal in several contexts.</param>
+        /// <param name="description">
+        /// The description of this Action, used to explain this goal in several contexts.
+        /// </param>
+        protected internal Action(string name, string? description = null)
```
joachimdekker: I am not a fan of a required name. If the name is for debugging purposes, I would say it must be possible to not name the action.

If one wants to change the name of the action, it can do so via the metadata:

```cs
Action action = new(() => {Console.WriteLine("It's gonna be Legen... wait for it... DARY");});
action.Metadata.Name = "Barney's Action";
action.Metadata.Description = "FRIENDSHIP OVER";
```

Furthermore, if we want to add something to the metadata, we would need to alter all constructors. Therefore, I would propose that we don't ask for the name and description, but for the user to supply the metadata object themselves:

```cs
Action action = new
(
    () => Console.WriteLine("It's gonna be Legen... wait for it... DARY"), 
    new Metadata(name: "Barney's Action", description: "FRIENDSHIP OVER")
);
```

joachimdekker (changes requested):
Good addition! I have left a few comments on the structure and future of the metadata. However, I just wanted to say I love the tests you have written!

Aplib.Core/Intent/Actions/Action.cs
Laura-Berkers: Yes I completely agree, that sounds way better. I will change it.

Laura-Berkers: About the required name though. I think there are thee ways to make the name not required.

1. Make the Metadata field an optional.
2. Make the Name field in Metadata optional.
3. Add a GUID to the metadata object.

I don't like really like the first option but it is possible.

I think the second option is not great, because then an object may have a Metadata attached that is non-null, but only containing null data.

I kind of like the last option, because it will also later on give us an easy way to distinguish components with certainty. And GUIDs are built into C# so it is easy.

joachimdekker: Great options! I think the second one is the way to go. 

You could create a new Metadata object everytime you instance the class using default properties. You could also have a constructor which can have an optional parameter `metadata` which if it is null, will be instanced to an empty Metadata class. This way, we always have a way to provide metadata in the constructor, but also outside the constructor.

joachimdekker: We could add a GUID to the metadata object. That is actually a great idea, and can be used as a sort of 'default' name. However, do make the GUID read-only.

JensSteenmetz:
I changed the PR name, so I disabled auto-merge. I think it's better if you enable it again yourself, as it is your PR.

Aplib.Core/Intent/Tactics/AnyOfTactic.cs
```diff
@@ -31,9 +35,13 @@ public AnyOfTactic(params Tactic[] subTactics)
         /// <summary>
         /// Initializes a new instance of the <see cref="AnyOfTactic"/> class with the specified sub-tactics and guard condition.
         /// </summary>
-        /// <param name="subTactics">The list of sub-tactics.</param>
         /// <param name="guard">The guard condition.</param>
-        public AnyOfTactic(Func<bool> guard, params Tactic[] subTactics) : this(subTactics) => Guard = guard;
+        /// <param name="metadata">
+        /// Metadata about this tactic, used to quickly display the tactic in several contexts.
+        /// </param>
+        /// <param name="subTactics">The list of sub-tactics.</param>
```
JensSteenmetz, line 42:
```suggestion
        /// <param name="subTactics">The list of subtactics.</param>
```

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -11,17 +11,25 @@ public class FirstOfTactic : AnyOfTactic
...
         /// <summary>
         /// Initializes a new instance of the <see cref="FirstOfTactic"/> class with the specified sub-tactics and guard condition.
         /// </summary>
-        /// <param name="subTactics">The list of sub-tactics.</param>
         /// <param name="guard">The guard condition.</param>
-        public FirstOfTactic(Func<bool> guard, params Tactic[] subTactics) : base(guard, subTactics)
+        /// <param name="metadata">
+        /// Metadata about this tactic, used to quickly display the tactic in several contexts.
+        /// </param>
+        /// <param name="subTactics">The list of sub-tactics.</param>
```
JensSteenmetz, line 30:
```suggestion
        /// <param name="subTactics">The list of subtactics.</param>
```

JensSteenmetz: Could you actually ctrl+f sub-tactics and change it to subtactics everywhere in the solution?

JensSteenmetz (changes requested):
Nice PR! There's a small spelling mistake. Apart from that and your discussion with Joachim about name/GUID, looks good to me!

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
Laura-Berkers: Hmm most of those lines are not currently touched by this PR. So I think if this needs to be fixed, dogmatically, it should be fixed in a different PR.

JensSteenmetz: Yeah I thought that might be the case

JensSteenmetz (dismissed):
Lgtm

Aplib.Core/Metadata.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// Store information about a component which may be useful for debugging or logging or general overviews.
+        /// </summary>
+        /// <param name="name">The name used to display the component.</param>
+        /// <param name="description">The description used to describe the component.</param>
+        public Metadata(string? name = null, string? description = null)
+        {
+            Id = Guid.NewGuid();
+            Name = name;
+            Description = description;
+        }
```
joachimdekker, line 35: Could you add an (internal) constructor which also takes an ID. This will be better when testing with metadata.

Aplib.Tests/Core/MetadataTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        // Arrange
+        const string name = "A program without bugs";
+        const string description = "This goal is unreachable \ud83d\ude22";
+
+        // Act
+        Metadata data = new(name, description);
+
+        // Assert
+        data.Should().NotBeNull();
+        data.Id.Should().NotBeEmpty();
```
joachimdekker, line 20: This is a weak test, as `Guid` is a struct and will always not be empty (the default value). As such, it would be better if you check if the guid is not `Guid.Empty`.

Aplib.Tests/Core/MetadataTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        data.Name.Should().BeNull();
+        data.Description.Should().Be(description);
+    }
+
+    [Fact]
+    public void Metadata_MultipleInstances_HaveUniqueIds()
```
joachimdekker: This is nitpicky, but there _exists_ a chance that this test will fail. Should this really be tested? I think we can rely on the implementation of `Guid.New()` for this.

Aplib.Tests/Core/MetadataTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        data.Name.Should().BeNull();
+        data.Description.Should().BeNull();
+
+        copy.Should().NotBeNull();
+        copy.Id.Should().NotBeEmpty();
+        copy.Name.Should().BeNull();
+        copy.Description.Should().BeNull();
+
+        data.Id.Should().Be(copy.Id);
+    }
```
joachimdekker: Is this test necessary? There is no copying, only reassigning the object to a different variable. The variables will both point to the same location in memory, so it is trivial that everything would be the same.

joachimdekker (changes requested):
Looks fine, I had a few nitpicks in the tests.

Aplib.Tests/Core/MetadataTests.cs
Laura-Berkers: That is actually what this does :) <https://fluentassertions.com/guids/>

joachimdekker: Well, TIL. Thanks 😄

JensSteenmetz (dismissed):
lgtm

joachimdekker (dismissed):
LGTM!

Tboefijn (approved):

joachimdekker (approved):
Lgtm

PR #24, opened 2024-03-26
feat: add MemoryBelief

JoenvandeVorle:
Add the belief which can store values, also from previous frames. The implementation supports remembering the last frame, but also the last _n_ frames. 
Includes an implementation of a CircularArray class to store the memories in a Queue-like datastructure that also allows indexing.


Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        /// <summary>
+        /// Generates/updates the resource if the shouldUpdate condition is satisfied.
+        /// Also stores the previous observation in memory.
+        /// </summary>
+        public override void UpdateBelief()
+        {
+            // We use the implicit conversion to TObservation to store the observation
+            if (_shouldUpdate()) _memorizedObservations.Put(this);
```
JensSteenmetz: Is this desired behavior? Should the memory not be added regardless of whether the observation is updated? I would expect to store a memory fragment every time I call UpdateBelief. For example, when I call GetMostRecentMemory() for two distinct memory beliefs, I expect both to be information about the _same_ previous game state. 

That said, the current implementation could be an idea for another specialized belief type? Or maybe a flag in the MemoryBelief about what exactly is memorized by this belief (i.e., previous observations, or previous _distinct_ observations, or maybe even another way to store memory). For now this might be out of scope.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// Gets the memorized resource at a specific index
+        /// </summary>
+        /// <returns> The memory of the resource at the specified index.</returns>
+        public TObservation GetMemoryAt(int index) => _memorizedObservations[index];
+
+        /// <summary>
+        /// Gets all the memorized resources
+        /// </summary>
+        /// <returns> A list of all the memorized resources.</returns>
+        public TObservation[] GetAllMemories() => _memorizedObservations.ToArray();
```
JensSteenmetz: What if the _memorizedObservations array has not been filled yet? Do we want to return an array with empty spaces? Or return a smaller array with only actual memorized observations? Again, for CircularArray in isolation I feel like the current implementation is what it it is supposed to do, but I'm not convinced for a _memory array_.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// Gets the most recently memorized resource
+        /// </summary>
+        /// <returns> The most recent memory of the resource.</returns>
+        public TObservation GetMostRecentMemory() => _memorizedObservations.GetLast();
+
+        /// <summary>
+        /// Gets the memorized resource at a specific index
+        /// </summary>
+        /// <returns> The memory of the resource at the specified index.</returns>
+        public TObservation GetMemoryAt(int index) => _memorizedObservations[index];
```
JensSteenmetz: Is this indexing behavior intuitive? I _think_ it's good that it cannot throw an out-of-range exception.

Assuming indexing the memory array cannot throw an out-of-range exception (as of now), I have a remark. I think that when considering CircularArray _in isolation_ it makes sense that a CircularArray wraps around when 'getting' a value, but for _memory_ it might be more logical to return the last memory item instead of wrapping to the most recent memory item again. When requesting the 100th memory item when there are 99 memory items, I would _not_ expect to get the most recent memory item, but that's what it does now.

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,96 @@
+using System;
+
+namespace Aplib.Core
+{
+    /// <summary>
+    /// An array that wraps around when it reaches its end.
+    /// Functionally works as a queue with indexing.
+    /// </summary>
+    public class CircularArray<T>
```
JensSteenmetz, line 7: I would suggest giving CircularArray a Length property (which is assigned in the constructor). This also allows _array.Length to be replaced by Length everywhere in this class.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        /// <summary>
+        /// Gets the most recently memorized resource
+        /// </summary>
+        /// <returns> The most recent memory of the resource.</returns>
+        public TObservation GetMostRecentMemory() => _memorizedObservations.GetLast();
+
+        /// <summary>
+        /// Gets the memorized resource at a specific index
```
JensSteenmetz: I would specify that a higher index means 'further back in time' (also note your punctuation in all your XML comments).
Edit: I just realized it's the other way around. I hope that reinforces my point XD. Also maybe switch that around? See my comment in CircularArray.cs.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// </summary>
+        /// <returns> The memory of the resource at the specified index.</returns>
+        public TObservation GetMemoryAt(int index) => _memorizedObservations[index];
+
+        /// <summary>
+        /// Gets all the memorized resources
+        /// </summary>
+        /// <returns> A list of all the memorized resources.</returns>
```
JensSteenmetz: Array*, not a list. Also, I would comment on the order of the memorized observations (oldest first, newest last, for the _current_ implementation).

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,81 @@
+using System;
+using System.Collections.Generic;
+using Aplib.Core;
+
+namespace Aplib.Core.Belief
+{
+	/// <summary>
+    /// The <see cref="MemoryBelief{TReference, TResource}"/> class represents a belief with "memory".
+    /// Some <i>object reference</i> is used to generate/update a <i>resource</i>
```
JensSteenmetz: I think you copied some XML comments from the `Belief` class; there were some changes to that. Please check what changed since you copied it. For example, _resource_ should be renamed to _observation_ everywhere, `TResource` should be `TObservation`, and I think there were some more changes.

Aplib.Tests/Core/Belief/CircularArrayTests.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    /// <summary>
+    /// Given a CircularArray instance,
+    /// When an element is put into the array,
+    /// The array should wrap around when it reaches its end.
+    /// (i.e., the first element should become the previous second element, 
+    /// and the last element should be the new element)
+    /// </summary>
+    [Fact]
+    public void Put_ArrayIsFull_UpdatesHead()
```
JensSteenmetz:
```suggestion
    public void Put_ArrayIsFull_WrapsAround() 
```

Aplib.Tests/Core/Belief/CircularArrayTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /// <summary>
+    /// Given a CircularArray instance,
+    /// When the head is updated,
+    /// Setting an element should set the correct index 
+    /// even if the head is not at the start of the array.
+    /// </summary>
+    [Fact]
+    public void Set_HeadIsUpdated_SetsCorrectIndex()
```
JensSteenmetz:
```suggestion
    public void Put_HeadIsUpdated_SetsCorrectIndex()
```

Aplib.Tests/Core/Belief/CircularArrayTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        // Assert
+        Assert.Equal(2, circularArray[0]);
+        Assert.Equal(3, circularArray[1]);
+        Assert.Equal(4, circularArray[2]);
+    }
+
+    /// <summary>
+    /// Given a CircularArray instance,
+    /// When the head is updated,
+    /// Setting an element should set the correct index
```
JensSteenmetz:
```suggestion
    /// Putting an element should set the correct index 
```

Aplib.Tests/Core/Belief/CircularArrayTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    [Fact]
+    public void Set_HeadIsUpdated_SetsCorrectIndex()
+    {
+        // Arrange
+        CircularArray<int> circularArray = new(3);
+
+        // Act
+        circularArray[1] = 4;
+        circularArray.Put(6);
+        circularArray[1] = 5;
```
JensSteenmetz: Maybe add comments on what exactly is happening here? For example
```suggestion
        // circularArray.ToArray() == [0, 0, 0]
        circularArray[1] = 4;
        // circularArray.ToArray() == [0, 4, 0]
        circularArray.Put(6);
        // circularArray.ToArray() == [4, 0, 6]
        circularArray[1] = 5;
        // circularArray.ToArray() == [4, 5, 6]
```

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,96 @@
...
+        private void IncrementHead()
+        {
+            _head = (_head + 1) % _array.Length;
+        }
+
+        /// <summary>
+        /// Puts an element at the end of the array.
+        /// </summary>
+        /// <param name="value">The element to add to the array</param>
+        public void Put(T value)
```
JensSteenmetz, line 61: From the name alone, I would expect this to put an element at the _start_ of the array. I see a CircularArray more like a queue, but with indexing of elements and a fixed maximum number of elements (before they 'fall out' the other end of the queue).

JensSteenmetz (changes requested):
Nice code and great tests. I left a few remarks on functionality and style.

Aplib.Core/Belief/MemoryBelief.cs
JoenvandeVorle: I think that makes more sense yeah, I'll change it to be that way. Maybe we can add a DistinctMemoryBelief later if we need it, it won't be much work

Aplib.Core/Belief/MemoryBelief.cs
JoenvandeVorle: I'll make it a shorter array for memory

Aplib.Core/Belief/Belief.cs
```diff
@@ -28,7 +28,7 @@ public class Belief<TReference, TObservation> : IBelief
         /// <summary>
         /// A condition on when the observation should be updated.
         /// </summary>
-        private readonly Func<bool> _shouldUpdate = () => true;
+        protected readonly Func<bool> _shouldUpdate = () => true;
```
JensSteenmetz:
```suggestion
        private readonly Func<bool> _shouldUpdate = () => true;
```

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,104 @@
...
+        private readonly CircularArray<TObservation> _memorizedObservations;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MemoryBelief{TReference, TResource}"/> class.
+        /// </summary>
+        /// <param name="reference">The reference used to generate/update the resource.</param>
+        /// <param name="getResourceFromReference">A function that takes a reference and generates/updates a resource.</param>
+        /// <param name="framesToRemember">The number of frames to remember back.</param>
+        public MemoryBelief(TReference reference, Func<TReference, TObservation> getResourceFromReference, int framesToRemember)
+            : base(reference, getResourceFromReference, () => true)
```
JensSteenmetz: Can remove the lambda.
```suggestion
        public MemoryBelief(TReference reference, Func<TReference, TObservation> getObservationFromReference, int framesToRemember)
            : base(reference, getObservationFromReference)
```

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// If the index is out of bounds, returns the closest element that is in bounds.
+        /// </summary>
+        /// <returns> The memory of the resource at the specified index.</returns>
+        public TObservation GetMemoryAt(int index)
+        {
+            if (index < 0) return GetMostRecentMemory();
+            if (index >= _memorizedObservations.Length)
+                return _memorizedObservations[_memorizedObservations.Length - 1];
+            return _memorizedObservations[index];
+        }
```
JensSteenmetz: Actually maybe give the user the option to clamp? On second thought an out-of-range exception by default seems like the 'safer' approach.
```suggestion
        public TObservation GetMemoryAt(int index, bool clamp = false)
        {
            int lastMemoryIndex = _memorizedObservations.Length - 1;
            if (clamp)
                index = Math.Clamp(index, 0, lastMemoryIndex);
            else if (index < 0 || index > lastMemoryIndex)
                throw new IndexOutOfRangeException();
            return _memorizedObservations[index];
        }
```
Also, I feel like you would like to return the _actual_ memory at index i. So when you have room for 5 memories and request the memory at index 1, but there is only a saved memory at index 0, then maybe return the memory at index 0 instead of the one at index 4 (if clamped).

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// If not all slots are filled, returns a smaller array.
+        /// </summary>
+        /// <returns> An array of all the memorized resources.</returns>
+        public TObservation[] GetAllMemories()
+        {
+            // Keep track of last non-default index in case of empty slots in the middle
+            int lastNonDefaultIndex = -1;
+            for (int i = 0; i < _memorizedObservations.Length; i++)
+            {
+                if (!_memorizedObservations[i]!.Equals(default(TObservation)))
```
JensSteenmetz: This could create issues when you store a default value as a memory. Maybe store the number of actual observations somewhere? Possibly slightly adjust the CircularArray data structure, maybe to something like an ExposedQueue (so it would function as a regular Queue where you can only put elements at the start, but can peek inside the Queue at any time)?

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// An array that wraps around when it reaches its end.
+    /// Functionally works sort of like a queue with indexing.
+    /// </summary>
+    public class CircularArray<T>
+    {
+
+        /// <summary>
+        /// The length of the array.
+        /// </summary>
+        public int Length { get; private set; }
```
JensSteenmetz, line 12: To accommodate the above conversations, change this to MaxLength, and add a Length component that indicates the actual length of the ExposedQueue?

JensSteenmetz (changes requested):
Added some more suggestions

Aplib.Core/Belief/MemoryBelief.cs
JensSteenmetz: There are still many of these throughout the document. Also some summaries were made more elaborative in Belief, I'd to more or less copy those for MemoryBelief

Aplib.Core/CircularArray.cs
JensSteenmetz: Or rename to MaxCount and Count and implement ICollection and IEnumerable. Not really necessary for us, so you could leave it, but would be nice.

Aplib.Core/Belief/MemoryBelief.cs
JoenvandeVorle: TODO: fix in later ticket

Aplib.Core/Belief/MemoryBelief.cs
JoenvandeVorle: TODO: fix in later ticket

Aplib.Core/CircularArray.cs
JoenvandeVorle: TODO: fix in later ticket

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -70,35 +76,40 @@ public override void UpdateBelief()
...
+            //     {
+            //         lastNonDefaultIndex = i;
+            //     }
+            // }
+            // if (lastNonDefaultIndex == -1) return Array.Empty<TObservation>();
 
-            TObservation[] memories = _memorizedObservations.ToArray(0, lastNonDefaultIndex);
-            return memories;
+            // TObservation[] memories = _memorizedObservations.ToArray(0, lastNonDefaultIndex);
+            // return memories;
```
JensSteenmetz: I would remove this commented-out code block.

JensSteenmetz (changes requested):

JensSteenmetz:
**New ticket:**
`CircularArray`: Iterating over the Array until a default value is reached (in `GetAllMemories`) could create issues when you store a default value as a memory. Maybe store the number of actual observations somewhere in CircularArray? Probably even change the CircularArray data structure, to something like an `ExposedQueue` (so it would function as a regular Queue where you can only put elements at the start, but can peek inside the Queue at any time)? Change Length to MaxCount, and add a Count component that indicates the actual length of the ExposedQueue? Implement ICollection and IEnumerable.
`GetMemoryAt` in `MemoryBelief`: I feel like you would like to return the actual memory at index i. So when you have room for 5 memories and request the memory at index 1, but there is only a saved memory at index 0, then maybe return the memory at index 0 instead of the one at index 4 (if clamped).

JensSteenmetz (dismissed):
STBC (seems to be correct)!

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+
+namespace Aplib.Core.Belief
+{
+    /// <summary>
+    /// The <see cref="MemoryBelief{TReference, TObservation}"/> class represents the agent's belief of a single object,
+    /// but with additional "memory" of previous observations.
+    /// Some <i>object reference</i> is used to generate/update a <i>observation</i>
```
QuakeEye:
```suggestion
    /// Some <i>object reference</i> is used to generate/update an <i>observation</i> 
```

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,106 @@
+
+namespace Aplib.Core
```
QuakeEye:
```suggestion
namespace Aplib.Core
```

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,106 @@
...
+{
+    /// <summary>
+    /// An array that wraps around when it reaches its end.
+    /// Functionally works sort of like a queue with indexing.
```
QuakeEye:
```suggestion
    /// Functionally works like a queue with indexing.
```

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// </summary>
+    public class CircularArray<T>
+    {
+
+        /// <summary>
```
QuakeEye:
```suggestion
        /// <summary>
```

Aplib.Core/CircularArray.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <param name="end">The end index of the range to copy.</param>
+        /// <returns>The circular array as a normal array</returns>
+        public T[] ToArray(int start = 0, int end = -1)
+        {
+            end = end == -1 ? Length - 1 : end;
+            T[] result = new T[end - start + 1];
+            for (int i = 0; i < result.Length; i++)
+            {
+                result[i] = this[start + i];
+            }
```
QuakeEye: What are our rules on this? Are curly brackets required here? As it is a oneliner.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// Gets the most recently memorized observation.
+        /// </summary>
+        /// <returns> The most recent memory of the observation.</returns>
+        public TObservation GetMostRecentMemory() => _memorizedObservations.GetFirst();
+
+        /// <summary>
+        /// Gets the memorized observation at a specific index.
+        /// A higher index means a memory further back in time.
+        /// If the index is out of bounds, returns the closest element that is in bounds.
```
QuakeEye: I assume this means the closest element that is in bounds, closest to the specified index? It could also be interpreted as the closest element to the current frame/now/0 so specifying this might be nice.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// The first element is the newest memory.
+        /// </summary>
+        /// <returns> An array of all the memorized observations.</returns>
+        public TObservation[] GetAllMemories()
+        {
+            // For now, we return the entire array, but with empty elements for the unused slots
+            return _memorizedObservations.ToArray();
+        }
+    }
+}
```
QuakeEye:
```suggestion
}

```

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        /// <summary>
+        /// Gets all the memorized observations.
+        /// The first element is the newest memory.
+        /// </summary>
+        /// <returns> An array of all the memorized observations.</returns>
+        public TObservation[] GetAllMemories()
+        {
+            // For now, we return the entire array, but with empty elements for the unused slots
```
QuakeEye, line 96: For now? This will change later? Does a `TODO:` tag need to be added? Has a Jira ticket been created/has a backlog item been added to the backlog items channel?

QuakeEye:
Could you please fix the spacing in `IBeliefSet` as part of this PR, so no new PR needs to be created.

Aplib.Tests/Core/Belief/CircularArrayTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        CircularArray<int> circularArray = new([1, 2, 3, 4, 5]);
+
+        // Act
+        circularArray.Put(0);
+        int[] array = circularArray.ToArray(1, 3);
+
+        // Assert
+        Assert.Equal([1, 2, 3], array);
+    }
+}
```
QuakeEye:
```suggestion
}

```

Aplib.Tests/Core/Belief/MemoryBeliefTests.cs
```diff
@@ -0,0 +1,94 @@
+using Aplib.Core;
```
QuakeEye: Unused import

QuakeEye (changes requested):

Aplib.Core/Belief/MemoryBelief.cs
JoenvandeVorle: Ticket has been made. I cant write `TODO ` because Sonar gets mad :(

Aplib.Core/CircularArray.cs
JoenvandeVorle: Oneliners are okay here apparently!

Aplib.Core/Belief/MemoryBelief.cs
QuakeEye: That was resolved, should add for now anyway, or make some other way to denote it needing change or being not implemented

JensSteenmetz (dismissed):
SAIYAM (seems adequate if you ask me)

QuakeEye (approved):

JensSteenmetz (approved):
saaa (still amazing as always)

PR #32, opened 2024-04-04
fix: update heuristics properly

SilasPeters:
## Description

The `_currentHeuristics` field of a Goal is never updated, as the property only set its backing field when the backing field is `null`, which it only is in the first frame.

This PR removes the backing field completely, and instead recalculates the heuristics every time the property is invoked.

## Testability

See the new test I included. You can test this fix by confirming that a goal can be considered completed or failed, even after multiple frames of being unfinished.

## Checklist

### DIFF
--- a/Aplib.Tests/Desire/GoalTests.cs
+++ b/Aplib.Tests/Desire/GoalTests.cs
@@ -121,13 +121,14 @@ public class GoalTests
     public void Goal_WhereHeuristicsChange_UsesUpdatedHeuristics()
     {
         // Arrange
+        IBeliefSet beliefSetMock = Mock.Of<IBeliefSet>();
         bool shouldSucceed = false;
         Goal<IBeliefSet> goal = new TestGoalBuilder().WithHeuristicFunction(_ => shouldSucceed).Build();
 
         // Act
-        CompletionStatus stateBefore = goal.GetStatus(It.IsAny<IBeliefSet>());
+        CompletionStatus stateBefore = goal.GetStatus(beliefSetMock);
         shouldSucceed = true; // Make heuristic function return a different value on next invoke
-        CompletionStatus stateAfter = goal.GetStatus(It.IsAny<IBeliefSet>());
+        CompletionStatus stateAfter = goal.GetStatus(beliefSetMock);
 
         // Assert
         stateBefore.Should().Be(CompletionStatus.Unfinished);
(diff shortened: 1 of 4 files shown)


Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -112,6 +112,28 @@ public void Goal_WhereEvaluationIsPerformed_DoesNotInfluenceBelieveSet()
...
+        Goal<IBeliefSet> goal = new TestGoalBuilder().WithHeuristicFunction(_ => shouldSucceed).Build();
+
+        // Act
+        CompletionStatus stateBefore = goal.GetStatus(It.IsAny<IBeliefSet>());
+        shouldSucceed = true; // Make heuristic function return a different value on next invoke
+        CompletionStatus stateAfter = goal.GetStatus(It.IsAny<IBeliefSet>());
+
+        // Assert
+        stateBefore.Should().Be(CompletionStatus.Unfinished);
+        stateAfter.Should().Be(CompletionStatus.Success);
```
JensSteenmetz: Write Act without Moq methods.
```suggestion
        bool shouldSucceed = false;
        IBeliefSet beliefSet = Mock.Of<IBeliefSet>();
        Goal<IBeliefSet> goal = new TestGoalBuilder().WithHeuristicFunction(_ => shouldSucceed).Build();

        // Act
        CompletionStatus stateBefore = goal.GetStatus(beliefSet);
        shouldSucceed = true; // Make heuristic function return a different value on next invoke
        CompletionStatus stateAfter = goal.GetStatus(beliefSet);

        // Assert
        stateBefore.Should().Be(CompletionStatus.Unfinished);
        stateAfter.Should().Be(CompletionStatus.Success);
```

JensSteenmetz (changes requested):
Some small suggestion, looks good overall

SSelinn (dismissed):
LGTM

Aplib.Tests/Desire/GoalTests.cs
SilasPeters: Fixed!

SSelinn (approved):
LBTM (looks better to me)

Aplib.Tests/Desire/GoalTests.cs
```diff
@@ -121,13 +121,14 @@ public void Goal_WhereEvaluationIsPerformed_DoesNotInfluenceBelieveSet()
     public void Goal_WhereHeuristicsChange_UsesUpdatedHeuristics()
     {
         // Arrange
+        IBeliefSet beliefSetMock = Mock.Of<IBeliefSet>();
```
JensSteenmetz, line 124: I would simply name it beliefSet, but that's a nit.

JensSteenmetz (approved):
Looks good to me!

Aplib.Tests/Desire/GoalTests.cs
SilasPeters: Agree

PR #37, opened 2024-04-19
feat: add DesireSet interrupt behavior

JensSteenmetz:
## Description
This PR aims to add interruptions of 'side' GoalStructures to the DesireSet class using a stack-like data structure.

For the implementation, I encountered a tough question regarding the updating of the 'active GoalStructures stack'. Namely:

> "Given a GoalStructure stack with GoalStructures MAIN, A, B in that order (B on top). When the activation guard of A is satisfied in the new cycle, and the activation guard of B is *not*, then what should the new stack look like?"

My answer to this question was MAIN, B, A. My reasoning was that the newly _reactivated_ GoalStructures should have priority over old GoalStructures. I thought that removing B from the stack should be done through _failing B_ (this is currently not an aplib feature yet, but is planned to be added later). To accommodate the functionality of reactivating GoalStructures, I used a data structure that I would describe as an _activation stack_. Instead of a Push method, it has an Activate method that produces the desired behavior. I made it optimized for the reactivation of stack items in O(1) time, hence the name OptimizedActivationStack.

## Testability
Run tests.

## Checklist

### DIFF
--- a/Aplib.Core/DataStructures/CircularArray.cs
+++ b/Aplib.Core/DataStructures/CircularArray.cs
@@ -1,4 +1,4 @@
-namespace Aplib.Core
+namespace Aplib.Core.DataStructures
 {
     /// <summary>
     /// An array that wraps around when it reaches its end.
--- a/Aplib.Core/DataStructures/OptimizedActivationStack.cs
+++ b/Aplib.Core/DataStructures/OptimizedActivationStack.cs
@@ -1,29 +1,65 @@
-﻿using System;
-using System.Collections;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace Aplib.Core.DataStructures
 {
-    internal class OptimizedActivationStack<T>
+    /// <summary>
+    /// A stack that has a predefined set of items that can be <i>activated</i>
+    /// (i.e., pushed on top of the stack).
+    /// When an item that is already on the stack is activated, it is <i>reactivated</i>
+    /// (i.e., moved to the top of the stack).
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="OptimizedActivationStack{T}"/> allows for O(1) activation and
+    /// reactivation of an arbitrary stack item.
+    /// </remarks>
+    /// <typeparam name="T">The type of the items that are put on the stack.</typeparam>
+    public class OptimizedActivationStack<T>
     {
-        private ActivationStackItem? _top { get; set; }
+        /// <summary>
+        /// The top item on the stack.
+        /// </summary>
+        private ActivationStackItem? _top;
 
+        /// <summary>
+        /// Gets the activatable stack items.
+        /// </summary>
+        /// <remarks>
+        /// The stack items are exposed, since they should be accessible from the outside to
+        /// provide O(1) activation of a stack item with <see cref="Activate(ActivationStackItem)"/>.
+        /// </remarks>
         public IEnumerable<ActivationStackItem> ActivatableStackItems { get; }
 
+        /// <summary>
+        /// Gets the number of items that are currently activated (i.e., on the stack).
+        /// </summary>
         public int Count { get; private set; } = 0;
 
+        /// <summary>
+        /// Initializes an optimized activation stack with a set of activatable items.
+        /// </summary>
+        /// <param name="activatables">A set of activatable items that could be pushed on the stack.</param>
         public OptimizedActivationStack(T[] activatables)
         {
             // Setup the activatable stack items.
             ActivatableStackItems = activatables.Select(activatable => new ActivationStackItem(activatable, this));
         }
 
-        public void Push(ActivationStackItem item)
+        /// <summary>
+        /// Activates an item (i.e., pushes an item on top of the stack).
+        /// If the pushed item is already on the stack,
+        /// it is extracted from the stack before it is put on top again.
+        /// </summary>
+        /// <param name="item">
+        /// The stack item that is pushed on top of the stack (i.e., it is activated).
+        /// </param>
+        /// <exception cref="System.ArgumentException">
+        /// Thrown when an item is pushed that belongs to a different stack.
+        /// </exception>
+        public void Activate(ActivationStackItem item)
         {
             if (item.ActivationStack != this)
-                throw new ArgumentException("Cannot push an item that is not on this stack.");
+                throw new System.ArgumentException("Cannot push an item that is not on this stack.");
 
             // Handle the case when the stack is empty.
             if (_top is null)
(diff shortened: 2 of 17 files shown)


Aplib.Core/DataStructures/CircularArray.cs
```diff
@@ -1,4 +1,4 @@
-namespace Aplib.Core
+namespace Aplib.Core.DataStructures
```
joachimdekker: I am not a fan of `DataStructures` as this is a bit vague. Why don't we mimic `System` and call it `Aplib.Core.Collections`?

```suggestion
namespace Aplib.Core.Collections
```

Aplib.Core/DataStructures/OptimizedActivationStack.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        /// <summary>
+        /// Represents (i.e., encapsulates) an item on the activation stack.
+        /// </summary>
+        /// <remarks>
+        /// This class is public, because the whole stack item should be accessible from the outside to
+        /// provide O(1) activation of a stack item with <see cref="Activate(Item)"/>.
+        /// </remarks>
+        public class Item
```
joachimdekker: If we have extra functionality, I would propose creating a separate file for this class as it is not simply a dataclass for OptimizedActivationStack.

Furthermore, Since it is only for this stack, I would make the class `sealed`.

Aplib.Core/DataStructures/OptimizedActivationStack.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// <exception cref="System.ArgumentException">
+        /// Thrown when an item is pushed that belongs to a different stack.
+        /// </exception>
+        public void Activate(Item item)
+        {
+            if (item.ActivationStack != this)
+                throw new System.ArgumentException("Cannot push an item that is not an activatable of this activation stack.");
+
+            // Handle the case when the stack is empty.
+            if (_top is null)
```
joachimdekker: Wouldn't it be better semantically to test for `Count <= 0`?

```suggestion
            if (Count <= 0)
```

Aplib.Core/Desire/DesireSet.cs
```diff
@@ -12,22 +13,85 @@ public class DesireSet<TBeliefSet> : IDesireSet<TBeliefSet>
...
+        /// All active goal structures of the agent that still need to be finished are pushed on the stack.
+        /// </summary>
+        private readonly OptimizedActivationStack
+            <(IGoalStructure<TBeliefSet> goalStructure, System.Func<TBeliefSet, bool> guard)> _goalStructureStack;
+
+        /// <summary>
+        /// If there are no goal structures left to be completed, the status of this desire set is set to the main goal status.
+        /// </summary>
+        public CompletionStatus Status
+            => _goalStructureStack.Count == 0 ? _mainGoal.Status : CompletionStatus.Unfinished;
```
joachimdekker, line 31: The GoalstructureStack should always have one goal structure on it right?

Aplib.Core/Desire/DesireSet.cs
```diff
@@ -12,22 +13,85 @@ public class DesireSet<TBeliefSet> : IDesireSet<TBeliefSet>
...
+        private void ActivateRelevantGoalStructures(TBeliefSet beliefSet)
+        {
+            foreach (var goalStructureStackItem in _goalStructureStack.ActivatableStackItems)
+            {
+                System.Func<TBeliefSet, bool> guard = goalStructureStackItem.Data.guard;
+
+                if (!guard(beliefSet)) continue;
+
+                _goalStructureStack.Activate(goalStructureStackItem);
+            }
```
joachimdekker: This can also be done with a LINQ statement:
```suggestion
			// Filter all the items by their guard.
            IEnumerable itemsToActivate = _goalStructureStack.ActivatableStackItems
                .Where(item => item.Data.guard(beliefSet))
            
            // Activate the stacks of the interrupts.
            foreach (var item in itemsToActivate)
            {
                _goalStructureStack.Activate(item);
            }
```

I think it is nicer and better understandable this way.

joachimdekker (changes requested):
Excellent, as ever! I left a few comments. Mostly nitpicks.

Aplib.Core/DataStructures/OptimizedActivationStack.cs
JoenvandeVorle: I would also call this class slightly different, something like (Activation)StackItem

Aplib.Tests/DataStructures/OptimizedActivationStackTests.cs
```diff
@@ -0,0 +1,250 @@
+﻿using Aplib.Core.DataStructures;
+using FluentAssertions;
+using Moq;
+using Moq.Protected;
```
JoenvandeVorle: These imports are unused

JoenvandeVorle (changes requested):

JoenvandeVorle (changes requested):
left few small comments, lgtm otherwise

Aplib.Core/DataStructures/CircularArray.cs
JensSteenmetz: Agreed!

Aplib.Core/DataStructures/OptimizedActivationStack.cs
JensSteenmetz: I will make it sealed. However, I do not think it should be separated from this class. The Item class does not have any significance without the OptimizedActivationStack class. It is inherently part of the OptimizedActivationStack class, and should not exist on its own.

Aplib.Core/DataStructures/OptimizedActivationStack.cs
JensSteenmetz: Great idea

Aplib.Core/Desire/DesireSet.cs
JensSteenmetz: No, the main goal is treated as a normal goal and is deleted from the stack when it is completed. I think this makes sense semantically as well, since the agent needs to 'complete all active goals'. I.e., all goals on the stack. When no more goals are left, we can evaluate the status of the desire set (and thus the status of the agent).

Aplib.Core/Desire/DesireSet.cs
JensSteenmetz: Agreed. Nice solution

Aplib.Core/DataStructures/OptimizedActivationStack.cs
JensSteenmetz: Yeah, you are right

JoenvandeVorle (dismissed):
awimo!!

Aplib.Core/DataStructures/OptimizedActivationStack.cs
joachimdekker: You make a valid point. I don't wholeheartedly agree, but I am fine with it this way.

Aplib.Core/Desire/DesireSet.cs
joachimdekker: That makes a lot of sense! Thanks for explaining.

Aplib.Tests/Collections/OptimizedActivationStackTests.cs
```diff
@@ -0,0 +1,248 @@
...
+    public void ActivatableStackItems_WhenActivatablesAreGiven_ShouldContainEncapsulatedActivatables()
+    {
+        // Arrange
+        int[] activatables = [1, 1, 2, 3, 5, 8];
+        OptimizedActivationStack<int> activationStack = new(activatables);
+
+        // Act
+        IEnumerable<OptimizedActivationStack<int>.StackItem> activatableStackItems
+            = activationStack.ActivatableStackItems;
+        IEnumerable<int> items = activatableStackItems.Select(stackItem => stackItem.Data);
```
joachimdekker: This should belong to Assert as it is not what you are testing.

joachimdekker (dismissed):
Looks good. I left one nitpick for the tests.

Aplib.Core/DataStructures/OptimizedActivationStack.cs
JensSteenmetz: I'm curious about your counterarguments.

joachimdekker (approved):
Looks good!

JoenvandeVorle (approved):

PR #42, opened 2024-04-24
feat: Add Observation property to Belief

joachimdekker:
## Description

This PR proposes to add the observation property as described in APLIB-102. I have chosen not to add it to `IBelief` as this belief does not contain a generic type indicating what type the observation property must be.

## Testability

All changes are tested using Unit Tests.

## Checklist

## Acceptance Criteria
- The IBelief and all beliefs in Aplib.Core must correctly implement the observation property (per Liskov substitution)
- The observation must give back the object which the stored Observation type.

### DIFF
--- a/Aplib.Core/Belief/Belief.cs
+++ b/Aplib.Core/Belief/Belief.cs
@@ -4,7 +4,7 @@ namespace Aplib.Core.Belief
 {
     /// <summary>
     /// The <see cref="Belief{TReference, TObservation}"/> class represents the agent's belief of a single object.
-    /// Some <i>object reference</i> is used to generate/update an <i>observation</i> 
+    /// Some <i>object reference</i> is used to generate/update an <i>observation</i>
     /// (i.e., some piece of information of the game state as perceived by an agent).
     /// </summary>
     /// <remarks>
@@ -33,7 +33,11 @@ namespace Aplib.Core.Belief
         /// <summary>
         /// The observation represented by the belief (i.e., some piece of information of the game state as perceived by an agent).
         /// </summary>
-        private TObservation _observation;
+        public TObservation Observation
+        {
+            get;
+            protected set;
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Belief{TReference, TObservation}"/> class with an object reference,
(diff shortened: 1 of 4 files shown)


Aplib.Core/Belief/Belief.cs
```diff
@@ -56,26 +60,28 @@ public Belief(TReference reference, Func<TReference, TObservation> getObservatio
...
         }
 
         /// <summary>
-        /// Implicit conversion operator to allow a <see cref="Belief{TReference, TObservation}"/> object 
+        /// Implicit conversion operator to allow a <see cref="Belief{TReference, TObservation}"/> object
         /// to be used where a <typeparamref name="TObservation"/> is expected.
         /// </summary>
         /// <param name="belief">The <see cref="Belief{TReference, TObservation}"/> object to convert.</param>
-        public static implicit operator TObservation(Belief<TReference, TObservation> belief) => belief._observation;
+        public static implicit operator TObservation(Belief<TReference, TObservation> belief) => belief.Observation;
```
Laura-Berkers, line 76: Have you considered removing the implicit conversion, now that we have a public getter?

Laura-Berkers (commented):
See comment

Aplib.Core/Belief/Belief.cs
joachimdekker: This conversion is meant as a quality of life method (the public observation property as well). I don't think it is meant to be removed with the addition of the public property.

Laura-Berkers: Forcing the user to write `.Observation` results in clearer code, but also having the implicit conversion may be nice. I don't have a strong opinion, just wanted to check that it was considered.

Laura-Berkers (approved):
LGTM

Aplib.Core/Belief/Belief.cs
joachimdekker: You are absolutely right. However, removing the implicit conversion would result in a breaking change, meaning that all aplib tests wouldn't even compile. 

We must try to avoid a breaking change at all cost, but we could make it deprecated.

Laura-Berkers: True, I think we can keep it (un-deprecated), unless it ever starts causing issues with type checking.

Tboefijn (approved):
LGTM!

PR #45, opened 2024-04-27
fix: correct a documentation error in `Belief.cs`

Laura-Berkers:
## Description

This fixes an error in the documentation for the `reference` parameter of the `Belief(TReference, Func<TReference, TObservation>)` constructor. I also included two other small improvements.

## Testability

N/A

## Checklist

### DIFF
--- a/Aplib.Core/Belief/Belief.cs
+++ b/Aplib.Core/Belief/Belief.cs
@@ -8,8 +8,7 @@ namespace Aplib.Core.Belief
     /// (i.e., some piece of information of the game state as perceived by an agent).
     /// </summary>
     /// <remarks>
-    /// It implements the <see cref="IBelief"/> interface.
-    /// It supports implicit conversion to <typeparamref name="TObservation"/>.
+    /// Implicit conversion to <typeparamref name="TObservation"/> is supported, see <see cref="op_Implicit"/>.
     /// </remarks>
     /// <typeparam name="TReference">The type of the object reference used to generate/update the observation.</typeparam>
     /// <typeparam name="TObservation">The type of the observation that the belief represents.</typeparam>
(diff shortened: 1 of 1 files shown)


Aplib.Core/Belief/Belief.cs
```diff
@@ -8,8 +8,7 @@ namespace Aplib.Core.Belief
     /// (i.e., some piece of information of the game state as perceived by an agent).
     /// </summary>
     /// <remarks>
-    /// It implements the <see cref="IBelief"/> interface.
```
Laura-Berkers, line 11: The fact that it implements the `IBelief` interface is already visible in the docs. Extra documentation is dangerous because it might fall out of sync with the code if overlooked.

Aplib.Core/Belief/Belief.cs
```diff
@@ -70,8 +71,8 @@ public Belief(TReference reference, Func<TReference, TObservation> getObservatio
         public static implicit operator TObservation(Belief<TReference, TObservation> belief) => belief._observation;
 
         /// <summary>
-        /// Generates/updates the observation if the shouldUpdate condition is satisfied.
-        /// The observation is then updated by calling the getObservationFromReference function.
+        /// Generates/updates the observation if the shouldUpdate condition is satisfied. The observation is then
+        /// updated by calling the <c>getObservationFromReference</c> function.
```
joachimdekker, line 75: I wouldn't say something about the implementation, but just say that the update function provided by the user is called.

joachimdekker (changes requested):

JensSteenmetz (changes requested):
Why is this PR created specifically for the Belief class? There are more problems with documentation in the codebase, can we not fix these all at once?

Laura-Berkers:
> Why is this PR created specifically for the Belief class? There are more problems with documentation in the codebase, can we not fix these all at once?

You are right, I will make this a draft for now.

joachimdekker:
What is the current status of this PR? It has been over a month without changes. Has it already been fixed, or should more work been done? Which work-item is this related to?

joachimdekker:
Jens said this PR could be closed. Whenever necessary, it could be reopened again.

PR #49, opened 2024-05-02
feat: add SampledMemoryBelief

JensSteenmetz:
## Description
This PR aims to add the SampledMemoryBelief. This belief is needed for tasks that require a long history. We cannot store the history of all frames for a longer period, as that would impose memory issues. Hence, we sample the frames.
See https://zomsa.atlassian.net/browse/AP-25.

Also addresses an issue with the naming conventions. Hence the many file changes.

## Acceptance Criteria

- [x] The user should be able to make a belief that samples the observable every n frames.
- [x] The user should be able to choose between immediate feedback mode, where the belief gets updated every time (but not added to the samples until n frames have passed); or sample mode, where the latest sample is always returned.

## Testability
Run tests.

## Checklist


joachimdekker:
What was the issue exactly with the naming conventions? And why was this addressed in this PR, and not a seperate PR?

Aplib.Core/Belief/SampledMemoryBelief.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// Also stores the previous observation in memory every <c>sampleInterval</c>-th cycle.
+        /// </summary>
+        public override void UpdateBelief()
+        {
+            if (ShouldSampleMemory())
+            {
+                // We use the implicit conversion to TObservation to store the observation.
+                _memorizedObservations.Put(this);
+
+                if (_shouldUpdate()) UpdateObservation();
```
joachimdekker: This is the implementation of `MemoryBelief.UpdateBelief()`. To avoid code-duplication, we should use `base.UpdateBelief()`.

joachimdekker (changes requested):
Looks great, but I have one nitpick about the inheritance.

JensSteenmetz:
> What was the issue exactly with the naming conventions? And why was this addressed in this PR, and not a seperate PR?

The issue was that I could not adhere to our naming conventions when making a private property (not auto-property). I addressed it here, because I needed the new conventions. The problem was that I would have to wait for that PR. I could still split it up if you want? @joachimdekker

joachimdekker:
No need to split it up now. Just keep it in mind for future PRs 😄

joachimdekker (dismissed):
LGTM!

Aplib.Core/Belief/SampledMemoryBelief.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// Specifies how this sampled memory belief should be updated.
+        /// </summary>
+        private readonly UpdateMode _updateMode;
+
+        private int _counter = 0;
+
+        /// <summary>
+        /// The number of cycles that have passed since the last memory sample was stored.
+        /// </summary>
+        private int Counter
```
SilasPeters: Your comment says it all, you could rename this to `CycleCount`

Aplib.Core/Belief/SampledMemoryBelief.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            get => _counter;
+            set => _counter = value % _sampleInterval;
```
SilasPeters: I understand where you are coming from. The idea of the count being cyclic (modulo) is there, such that ShouldSample() can test whether is has cycled back to 0. However, this mindset influences both this Counter and the ShouldSample method. Both are intertwined and depended on each other, hoping someone does not change one of both in isolation - whilst this might happen as its not that intuitive. Lastly, its a bit of code coupling.

In sake of seperation of concern, and my long story from just now, I would suggest making the counter quite litterally a counter (who knows, might be useful in the future) and change the ShouldSample to modulo himself. If your idea was to represent how many times it was since the most recent cycle, put that idea in a dedicated method or field.

TL;DR one strong idea is spread over two fields, where one might accidentally change since the idea and coupling is not that obvious, and break the idea.

Aplib.Core/Intent/Tactics/Tactic.cs
```diff
@@ -14,6 +14,11 @@ namespace Aplib.Core.Intent.Tactics
     public abstract class Tactic<TBeliefSet> : ITactic<TBeliefSet>
         where TBeliefSet : IBeliefSet
     {
+        /// <summary>
+        /// Gets or sets the guard of the tactic.
+        /// </summary>
+        protected System.Func<TBeliefSet, bool> _guard = _ => true;
```
SilasPeters, line 20: Why remove the autoproperty? It's not readonly or anything.

SilasPeters (approved):
Quick review. Frankly, I skipped over the tests.

Aplib.Core/Belief/SampledMemoryBelief.cs
JensSteenmetz: While I kinda agree with your point, I have decided to make it a cyclic (modulo) counter for the following reason. When the integer reaches the maximum value, it loops around and in the modulo class this might skip a few steps in the counter. It's very unlikely to happen, but I would rather do it like this. I will call it ModuloCounter tho.

Aplib.Core/Belief/SampledMemoryBelief.cs
JensSteenmetz: I will call it ModuloCounter, as that seems to be the more commonly used name for this behavior.

Aplib.Core/Intent/Tactics/Tactic.cs
JensSteenmetz: It's not readonly because of a flaw in the code. I will suggest to open a backlog item for it, but it's caused by a problem that occurs at multiple places in the codebase.

Aplib.Core/Belief/SampledMemoryBelief.cs
SilasPeters: Oh you're right! Or BigInteger haha

SilasPeters (approved):

joachimdekker (approved):
Looks good.

PR #51, opened 2024-05-08
feat: replace CircularArray with ExposedQueue

JoenvandeVorle:
## Description

This changes the MemoryBelief to use an exposed queue instead of a circular array.

## Testability

Run the unit tests in `ExposedQueueTests` and `MemoryBeliefTests`.

## Checklist


Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        public int MaxCount { get; private set; }
+
+        /// <summary>
+        /// Actual number of elements in the array.
+        /// Limited to <see cref="MaxCount"/>.
+        /// </summary>
+        public int Count
+        {
+            get => _count;
+            private set => _count = value > MaxCount ? MaxCount : value;
```
joachimdekker: I'd say the count would never exceed MaxCount? If that is the case, and it gets caught here, we did something wrong. If the value is higher than MaxCount, an exception should be thrown instead of returning the max value.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// </summary>
+        /// <param name="array">An array to use as the circular array.</param>
+        /// <param name="count">The number of actual elements in the array.</param>
+        public ExposedQueue(T[] array, int count = -1)
+        {
+            MaxCount = array.Length;
+            _array = array;
+            _head = MaxCount - 1;
+            Count = count == -1 ? MaxCount : count;
+        }
```
joachimdekker: Multiple pointers:
* The user could provide a negative count, which will not be caught in an exception.
* I would propose using nullables here. The default count of `-1` is not ideal. The user could always provide a count of -1 themselves. This should throw an exception instead of setting the count to the max count. The ideal solution would be to use `MaxCount` as default value, but since it is computed at runtime, this is not possible.

```suggestion
        public ExposedQueue(T[] array, int? count = null)
        {
            MaxCount = array.Length;
            _array = array;
            _head = MaxCount - 1;
            Count = count ?? MaxCount;
        }
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        public T this[int index]
+        {
+            get => _array[(index + _head + 1) % MaxCount];
+            private set => _array[(index + _head + 1) % MaxCount] = value;
+        }
+
+        /// <summary>
+        /// Decrements the head of the array.
+        /// </summary>
+        private void DecrementHead() => _head = (_head - 1 + MaxCount) % MaxCount;
```
joachimdekker: This method is positioned wrong, and should be lower than all public methods.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Converts the ExposedQueue to an array.
+        /// </summary>
+        /// <param name="start">The start index of the range to convert.</param>
+        /// <param name="end">The end index of the range to convert.</param>
+        /// <returns>An array containing the elements within the specified range.</returns>
+        public T[] ToArray(int start, int end)
+        {
+            Console.WriteLine($"Start: {start}, End: {end}");
```
joachimdekker: Avoid leaving debug statements in code merging to `dev`

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        public bool Remove(T item)
+        {
+            for (int i = 0; i < Count; i++)
+                if (this[i]!.Equals(item))
+                {
+                    this[i] = default!;
+                    return true;
+                }
+            return false;
+        }
```
joachimdekker: Removing this element leaves a 'gap' in the array. Isn't this what this PR is supposed to solve? I am calling upon @JensSteenmetz to assist me here 😄

joachimdekker: Furthermore, it does not change the `count` property? It seems to me this does not work as intended.

joachimdekker (changes requested):
Left a few comments.

Aplib.Core/ExposedQueue.cs
JoenvandeVorle: I'm honestly not even sure if removing items at some random index is supposed to be possible, this is a kind of queue after all.

Aplib.Core/ExposedQueue.cs
JoenvandeVorle: I did it this way to avoid having to put `if (Count < MaxCount) Count++` in the `Put()` method. I can put it there, but then I would not put another check for an exception, since that would never occur.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <summary>
+        /// Gets the element at the head of the array.
+        /// </summary>
+        /// <returns>The element at the head of the array</returns>
+        public T GetHead() => _array[_head];
+
+        /// <summary>
+        /// Gets the first element of the array.
+        /// </summary>
+        /// <returns>The last element of the array</returns>
```
JensSteenmetz:
```suggestion
        /// <returns>The first element of the array.</returns>
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+            Count++;
+        }
+
+        /// <inheritdoc/>
+        public void Add(T item) => Put(item);
+
+        /// <summary>
+        /// Gets the element at the head of the array.
+        /// </summary>
+        /// <returns>The element at the head of the array</returns>
```
JensSteenmetz: Nit
```suggestion
        /// <returns>The element at the head of the array.</returns>
```

Aplib.Core/ExposedQueue.cs
JensSteenmetz: I would put it in the Put() method for clarity, but agree the exception is not necessary.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Decrements the head of the array.
+        /// </summary>
+        private void DecrementHead() => _head = (_head - 1 + MaxCount) % MaxCount;
+
+        /// <summary>
+        /// Puts an element at the start of the array.
+        /// </summary>
+        /// <param name="value">The element to add to the array</param>
```
JensSteenmetz: Nit
```suggestion
        /// <param name="value">The element to add to the array.</param>
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <summary>
+        /// Gets the element at the specified index.
+        /// </summary>
+        /// <param name="index">The index of the element to get.</param>
+        /// <returns>The element at the specified index.</returns>
+        public T this[int index]
+        {
+            get => _array[(index + _head + 1) % MaxCount];
+            private set => _array[(index + _head + 1) % MaxCount] = value;
+        }
```
JensSteenmetz: Could you throw an error when the specified index is higher than the Count? I don't think we want to make the index 'wrappable' like that.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Copies the circular array to an array.
+        /// The head should be the last element of the array.
+        /// Copies from start to end inclusive.
+        /// </summary>
+        /// <param name="array">The array to copy to."</param>
+        /// <param name="arrayIndex">The start index of the range to copy.</param>
+        /// <param name="endIndex">The end index of the range to copy.</param>
+        /// <returns>The circular array as a normal array</returns>
```
JensSteenmetz: Nit
```suggestion
        /// <returns>The circular array as a normal array.</returns>
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <returns>The last element of the array</returns>
+        public T GetFirst() => this[0];
+
+        /// <summary>
+        /// Copies the circular array to an array.
+        /// The head should be the last element of the array.
+        /// Copies from start to end inclusive.
+        /// </summary>
+        /// <param name="array">The array to copy to."</param>
+        /// <param name="arrayIndex">The start index of the range to copy.</param>
```
JensSteenmetz:
```suggestion
        /// <param name="startIndex">The start index of the range to copy.</param>
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+                throw new ArgumentException("Start index must be less than or equal to end index.");
+            if (array.Length < endIndex - arrayIndex + 1)
+                throw new ArgumentException("Array is too small to copy the range.");
+
+            for (int i = 0; i < endIndex - arrayIndex + 1; i++)
+                array[i] = this[arrayIndex + i];
+        }
+
+        /// <inheritdoc/>
+        public void CopyTo(T[] array, int arrayIndex) => CopyTo(array, arrayIndex, arrayIndex + Count - 1);
```
JensSteenmetz: Maybe nice to have a default value as well?
```suggestion
        public void CopyTo(T[] array, int startIndex = 0) => CopyTo(array, arrayIndex, arrayIndex + Count - 1);
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        private readonly T[] _array;
+        private int _head;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExposedQueue{T}"/> class.
+        /// </summary>
+        /// <param name="size">The size of the array.</param>
+        public ExposedQueue(int size)
+        {
+            MaxCount = size;
```
JensSteenmetz:
```suggestion
            MaxCount = size;
            Count = 0;
```

Aplib.Core/ExposedQueue.cs
JensSteenmetz: I agree, I don't think Remove is needed at all.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+                if (this[i]!.Equals(item))
+                    return true;
+            return false;
+        }
+
+        /// <inheritdoc/>
+        public bool Remove(T item)
+        {
+            for (int i = 0; i < Count; i++)
+                if (this[i]!.Equals(item))
```
JensSteenmetz: Is the null-forgiving operator necessary here?
```suggestion
                if (this[i].Equals(item))
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        /// <inheritdoc/>
+        public void Add(T item) => Put(item);
+
+        /// <summary>
+        /// Gets the element at the head of the array.
+        /// </summary>
+        /// <returns>The element at the head of the array</returns>
+        public T GetHead() => _array[_head];
```
JensSteenmetz: Could this be renamed to GetLast? 'Head' could be interpreted in various ways, while GetLast is less ambiguous semantically. This remark is just for the publically visible GetHead.
```suggestion
        public T GetLast() => _array[_head];
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <returns>The circular array as a normal array</returns>
+        public void CopyTo(T[] array, int arrayIndex, int endIndex)
+        {
+            if (arrayIndex < 0 || arrayIndex >= Count)
+                throw new ArgumentOutOfRangeException(nameof(arrayIndex));
+            if (endIndex < 0 || endIndex >= Count)
+                throw new ArgumentOutOfRangeException(nameof(endIndex));
+            if (arrayIndex > endIndex)
+                throw new ArgumentException("Start index must be less than or equal to end index.");
+            if (array.Length < endIndex - arrayIndex + 1)
```
JensSteenmetz: This one should be caught by the first two exceptions, right? Because one (or both) would have to be out of range.

Aplib.Tests/Belief/ExposedQueueTests.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        // Assert
+        Assert.Throws<ArgumentOutOfRangeException>(StartOutOfBounds);
+        Assert.Throws<ArgumentOutOfRangeException>(EndOutOfOunds);
+        Assert.Throws<ArgumentException>(StartAfterEnd);
+        Assert.Throws<ArgumentException>(TargetArrayTooSmall);
+    }
+
+    [Fact]
+    public void CopyTo_CopiesCorrectly()
```
JensSteenmetz: Nit
```suggestion
    public void CopyTo_WhenCalled_CopiesCorrectly()
```

Aplib.Tests/Belief/ExposedQueueTests.cs
```diff
@@ -0,0 +1,219 @@
...
+        // Act
+        int[] array = new int[3];
+        queue.CopyTo(array, 0);
+
+        // Assert
+        Assert.Equal(array, [3, 2, 1]);
+    }
+
+    [Fact]
+    public void ToArray_ReturnsCorrectArray()
```
JensSteenmetz: Nit
```suggestion
    public void ToArray_WhenCalled_ReturnsCorrectArray()
```

JensSteenmetz (changes requested):
Overall good work! Left a few comments.

Aplib.Core/ExposedQueue.cs
JensSteenmetz: It is, nvm

Aplib.Core/ExposedQueue.cs
JoenvandeVorle: Sonar says it should be arrayIndex, to match with the interface declaration.

Aplib.Core/ExposedQueue.cs
JoenvandeVorle: Same as other comment, sonar wants it to match the interface, which has no default parameter

Aplib.Core/ExposedQueue.cs
```diff
@@ -50,65 +44,76 @@ public ExposedQueue(int size)
...
-            Count = count == -1 ? MaxCount : count;
+            Count = count ?? MaxCount;
         }
 
         /// <summary>
-        /// Gets the element at the specified index.
+        /// Gets the element at the specified index. Throws an exception if the index is out of bounds.
         /// </summary>
         /// <param name="index">The index of the element to get.</param>
         /// <returns>The element at the specified index.</returns>
```
JensSteenmetz:
```suggestion
        /// <returns>The element at the specified index.</returns>
        /// <exception cref="ArgumentOutOfRangeException">
        /// Thrown when the index is out of range.
        /// </exception>
```

Aplib.Core/ExposedQueue.cs
```diff
@@ -50,65 +44,76 @@ public ExposedQueue(int size)
...
         public T this[int index]
         {
-            get => _array[(index + _head + 1) % MaxCount];
-            private set => _array[(index + _head + 1) % MaxCount] = value;
+            get 
+            {
+                if (index < 0 || index >= Count)
+                    throw new ArgumentOutOfRangeException(nameof(index));
```
JensSteenmetz: You already catch this when setting.
```suggestion
```

Aplib.Core/ExposedQueue.cs
JensSteenmetz: Ah, alright

Aplib.Core/ExposedQueue.cs
```diff
@@ -169,17 +171,15 @@ public bool Contains(T item)
...
-        }
+        /// <summary>
+        /// DOES NOT DO ANYTHING.
+        /// </summary>
+        /// <returns>False.</returns>
+        /// <remarks>
+        /// Method is there to comply with the ICollection interface. 
+        /// Does not actually do anything since this is a queue, and queues do not allow removal of specific elements.
+        /// </remarks>
+        public bool Remove(T item) => false;
```
JensSteenmetz: Ah, Remove is necessary for the ICollection. Didn't realize that. Maybe we should implement it then, but without leaving a gap. It's a bit of a hassle but I don't think it should do nothing actually. @joachimdekker what do you think?

JensSteenmetz (changes requested):
A few more comments

Aplib.Core/ExposedQueue.cs
joachimdekker: Incorrect. As ExposedQueue is a public class with a public indexer, there is a good reason to suspect people will index it wrong. Either way, it is a good practice to check this.

Aplib.Core/ExposedQueue.cs
joachimdekker: Yes. Leaving this with a false is bad practice, and will confuse people. You could better throw a NotImplementedException and discard the issue Sonar raises.

As for a better implementation, it should be possible to remove something from this queue if we make it an ICollection. However, don't think too hard about the complexity, we could just call `_array.Remove(item)` and update the `_head` and `Count` accordingly. This makes this method O(n), or worse, but that is allowed, since we have a Queue datastructure.

Aplib.Tests/Belief/ExposedQueueTests.cs
```diff
@@ -0,0 +1,231 @@
...
+    [Fact]
+    public void ExposedQueue_WhenInitializedWithArrayAndCount_ShouldThrowExceptions()
+    {
+        // Arrange
+        void CountExceedsLength() => _ = new ExposedQueue<int>([1, 2, 3], 4);
+        void NegativeCount() => _ = new ExposedQueue<int>([1, 2, 3], -1);
+
+        // Assert
+        Assert.Throws<ArgumentOutOfRangeException>(CountExceedsLength);
+        Assert.Throws<ArgumentOutOfRangeException>(NegativeCount);
```
joachimdekker: Please write two separate tests for this. Asserting multiple things in a test is a test smell, as it is difficult to quickly pinpoint what the point of failure is (was it a high count or a negative count that caused this test to fail?)

In general, I'd think multiple asserts per test are allowed if you need to assert multiple thing with the same setup. This method uses two act methods and thus do not comply.

Aplib.Tests/Belief/ExposedQueueTests.cs
```diff
@@ -0,0 +1,231 @@
...
+        ExposedQueue<int> queue = new(3);
+        int elem;
+
+        // Act
+        void AccessNegativeIndex() => elem = queue[-1];
+        void AccessIndexGreaterThanCount() => elem = queue[queue.Count];
+
+        // Assert
+        Assert.Throws<ArgumentOutOfRangeException>(AccessNegativeIndex);
+        Assert.Throws<ArgumentOutOfRangeException>(AccessIndexGreaterThanCount);
```
joachimdekker: Please write two separate tests for this. Asserting multiple things in a test is a test smell, as it is difficult to quickly pinpoint what the point of failure is (was it a negative index or an out of bound index that caused this test to fail?)

In general, I'd think multiple asserts per test are allowed if you need to assert multiple thing with the same setup. This method uses two act methods and thus do not comply.

Aplib.Tests/Belief/ExposedQueueTests.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        // Act
+        void StartOutOfBounds() => queue.CopyTo(new int[3], -1, 2);
+        void EndOutOfOunds() => queue.CopyTo(new int[3], 0, 3);
+        void StartAfterEnd() => queue.CopyTo(new int[3], 2, 1);
+
+        // Assert
+        Assert.Throws<ArgumentOutOfRangeException>(StartOutOfBounds);
+        Assert.Throws<ArgumentOutOfRangeException>(EndOutOfOunds);
+        Assert.Throws<ArgumentException>(StartAfterEnd);
```
joachimdekker: Please write three separate tests for this. Asserting multiple things in a test is a test smell, as it is difficult to quickly pinpoint what the point of failure is.

In general, I'd think multiple asserts per test are allowed if you need to assert multiple thing with the same setup. This method uses two act methods and thus do not comply.

Aplib.Tests/Belief/MemoryBeliefTests.cs
```diff
@@ -54,22 +55,22 @@ public void GetMemoryAt_WhenObservationIsUpdated_ShouldReturnObservationAtSpecif
...
-        list.Add(4);
-        belief.UpdateBelief();
+        void GetMemoryAtNegativeIndex() => belief.GetMemoryAt(-1);
+        void GetMemoryAtIndexGreaterThanCount() => belief.GetMemoryAt(3);
 
         // Assert
-        Assert.Equal(3, belief.GetMemoryAt(-1, true));
-        Assert.Equal(0, belief.GetMemoryAt(5, true));
+        Assert.Throws<ArgumentOutOfRangeException>(GetMemoryAtNegativeIndex);
+        Assert.Throws<ArgumentOutOfRangeException>(GetMemoryAtIndexGreaterThanCount);
```
joachimdekker: Please write two separate tests for this. Asserting multiple things in a test is a test smell, as it is difficult to quickly pinpoint what the point of failure is.

In general, I'd think multiple asserts per test are allowed if you need to assert multiple thing with the same setup (arrange and act). This method uses two act methods and thus do not comply.

Aplib.Core/ExposedQueue.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ExposedQueue{T}"/> class.
+        /// By default, assumes the array is filled.
+        /// </summary>
+        /// <param name="array">An array to use as the circular array.</param>
+        /// <param name="count">The number of actual elements in the array.</param>
+        public ExposedQueue(T[] array, int? count = null)
+        {
+            if (count > array.Length)
+                throw new ArgumentOutOfRangeException(nameof(count), "Count cannot exceed the length of the array.");
```
joachimdekker: Shouldn't you be allowed to initialize the queue with an array that is less than the maximal elements?

Furthermore, if the count is less than the length of the array, why are we still storing the whole array? Wouldn't it be better if we copied the array to another array with the length of `count`.

joachimdekker (changes requested):
Looks solid, but I left some architectural comments and found a couple of test smells.

Aplib.Core/ExposedQueue.cs
JoenvandeVorle: Yes, you can initialize the queue with an array that is less than the maximal elements. count here refers to the amount of *actual* elements in the array, not the maximal allowed amount. MaxCount is the array length, and Count is the amount of slots currently filled. 
If count is less than the array length, that simply means all slots have not been filled yet. I will try to make this more clear in the method summary.

Aplib.Core/Belief/MemoryBelief.cs
```diff
@@ -98,15 +98,12 @@ public override void UpdateBelief()
...
-        /// If the index is out of bounds, when clamped, returns the element closest to the index that is in bounds.
         /// </summary>
         /// <returns>The memory of the observation at the specified index.</returns>
-        public TObservation GetMemoryAt(int index, bool clamp = false)
+        public TObservation GetMemoryAt(int index)
         {
             int lastMemoryIndex = _memorizedObservations.Count;
-            if (clamp)
-                index = Math.Clamp(index, 0, lastMemoryIndex);
-            else if (index < 0 || index > lastMemoryIndex)
```
JensSteenmetz: Why did you remove this? I think it's alright to remove it, but then there must be a way of checking the current number of memorized observations. Right now there is no way of preventing an ArgumentOutOfRangeException, because we cannot check the allowed range of indices for the GetMemoryAt method.

Aplib.Core/ExposedQueue.cs
```diff
@@ -70,7 +84,7 @@ public ExposedQueue(T[] array, int? count = null)
                     throw new ArgumentOutOfRangeException(nameof(index));
                 return _array[(index + _head + 1) % MaxCount];
             }
-            private set
+            private set
```
JensSteenmetz: Nit
```suggestion
            private set
```

JensSteenmetz (changes requested):
Left one more remark

Aplib.Core/Belief/MemoryBelief.cs
JoenvandeVorle: I thought you said in an earlier comment that the clamping was unwanted, but maybe I'm hallucinating, can't find it again. Will re-add

JensSteenmetz (dismissed):
LGTM!!

joachimdekker (dismissed):
LGTM

JensSteenmetz (approved):
LGTM!!!

joachimdekker (approved):

PR #52, opened 2024-05-09
feat: add implicit lifting

SilasPeters:
## Description

Most notable changes, are:
 - All BDI cycle component now inherit `IDocumented`, which contains an `IMetadata`
 - In every file of an BDI cycle component, extension methods have been added in a partial class which create said component through lifting.
 - Every BDI cycle component which can be created by lifting a lower-order component, has implicit converters which perform the lifting.

Here is some example code of what has been made possible:

```csharp
public void Test()
{
  Action<TBeliefSet> a = new();
  Tactic<TBeliefSet> t1 = a.Lift();
  Tactic<TBeliefSet> t2 = a;

  Goal<TBeliefSet> g = new(t1, CommonHeuristicFunctions<TBeliefSet>.Completed());
  GoalStructure<TBeliefSet> gs = new Goal<TBeliefSet>(t1, CommonHeuristicFunctions<TBeliefSet>.Completed());
  DesireSet<TBeliefSet> d1 = g.Lift();
  DesireSet<TBeliefSet> d2 = g;
  DesireSet<TBeliefSet> d3 = gs.Lift();
  DesireSet<TBeliefSet> d4 = gs;

  IGoal<TBeliefSet> ginterface = new Goal<TBeliefSet>(t1, CommonHeuristicFunctions<TBeliefSet>.Completed());
  // DesireSet<TBeliefSet> d5 = ginterface; THIS WOULD NOT WORK, FOR AN INTERFACE CANNOT BE IMPLICITLY CONVERTED
  DesireSet<TBeliefSet> d6 = ginterface.Lift();
}

```

## Testability

Try the code above, and run all tests

## Checklist


Aplib.Core/IHasMetadata.cs
```diff
@@ -0,0 +1,16 @@
+namespace Aplib.Core
+{
+    /// <summary>
+    /// Represents an object that contains an instance of <see cref="IMetadata"/>.
+    /// </summary>
+    public interface IHasMetadata
```
joachimdekker: `IHasMetadata` is an unconventional way of creating interfaces. If we look at the C# standard library, we never see an interface contain 'has'. I would like to propose another name for the interface: `IDocumented`. This name clearly communicates the requirements for an object to have metadata, and conforms to the format many libraries use.

SilasPeters: Agreed

JensSteenmetz (changes requested):
I will review the rest later, but:
- You are now doing 2 tickets at once. See AP-122 and AP-105. I would suggest to split these. From the looks of it, you probably would want to do AP-122 first.
- Implicit lifting (AP-105) should *not* be breaking. Metadata in constructors (AP-122) will be.
- I would suggest to keep classes, so also `LiftingExtensionMethods`, in their own files.
- Why does Metadata need to have an interface? And why do we need a IHasMetaData interface? I don't see the benefit of it, so to me this looks like unnecessary/premature abstraction.
- Why can the code snippet we need to test not be made into a unit test? Or multiple unit tests?

SilasPeters:
>     * You are now doing 2 tickets at once. See AP-122 and AP-105. I would suggest to split these. From the looks of it, you probably would want to do AP-122 first.

If I were to remove any changes to metadata, I would create only halve an implementation of lifting. I suggest I await the metadata changes then, if more people disagree.

>     * I would suggest to keep classes, so also `LiftingExtensionMethods`, in their own files.

But you agreed to this in the dominos :cry: . But yeah, it's probably better.

>     * Why does Metadata need to have an interface? And why do we need a IHasMetaData interface? I don't see the benefit of it, so to me this looks like unnecessary/premature abstraction.

In case someone wants to define his own Metadata. If we don't do this, we would not allow it, for lifting would require a concrete `Metadata` then.

JensSteenmetz:
> If I were to remove any changes to metadata, I would create only halve an implementation of lifting. I suggest I await the metadata changes then, if more people disagree.

I would do it the other way around; fix the metadata first, then fix the lifting after. Lifting should be a smooth addition when the rest of the code is fixed.

> But you agreed to this in the dominos 😢 .

I think I misunderstood then. I agreed with the LiftingExtensionMethods class. I also agreed with using partial classes in our entire code base, because that makes our classes extendable for others as well. However, I didn't realize you meant putting LiftingExtensionMethods in the same class.

> In case someone wants to define his own Metadata. If we don't do this, we would not allow it, for lifting would require a concrete `Metadata` then.

For IHasMetadata: I still don't see how that is necessary, even if someone wants to define his own metadata.
For IMetadata: Alright, but then IMetadata should have different members I guess? I might not want to give my objects a description. Maybe I just want to give them a name, for example. Or just an ID. Even then, however, I don't really see much of a benefit, since it completely depends on how we will go about *using* the metadata. I would like to be hesitant with abstraction, as it can also create *more* work later if it's the 'wrong' abstraction for our use case.

joachimdekker:
What is the status of this PR, it has been 3 weeks.

SilasPeters:
> What is the status of this PR, it has been 3 weeks.

I am awaiting the constructor fix, as discussed with others

Aplib.Core/LiftingExtensionMethods.cs
```diff
@@ -0,0 +1,121 @@
...
+            where TBeliefSet : IBeliefSet
+        {
+            string mostDerivedActionType = action.GetType().Name;
+
+            IMetadata metadata = action is not IDocumented actionWithMetadata
+                ? new Metadata($"Lifted {mostDerivedActionType}")
+                : new Metadata(!string.IsNullOrWhiteSpace(actionWithMetadata.Metadata.Name)
+                    ? $"[Lifted {mostDerivedActionType}] {actionWithMetadata.Metadata.Name}"
+                    : $"Lifted {mostDerivedActionType}",
+                    actionWithMetadata.Metadata.Description);
```
JensSteenmetz: Why do we not just lift without metadata? Or optional new metadata? So:
```cs
public static PrimitiveTactic<TBeliefSet> Lift<TBeliefSet>(this IAction<TBeliefSet> action)
            where TBeliefSet : IBeliefSet
{
...
}


public static PrimitiveTactic<TBeliefSet> Lift<TBeliefSet>(this IAction<TBeliefSet> action, Metadata metadata)
            where TBeliefSet : IBeliefSet
{
...
}
```

Same comment on all other extension methods.

Aplib.Core/Desire/DesireSets/DesireSet.cs
```diff
@@ -93,5 +108,20 @@ public void Update(TBeliefSet beliefSet)
                 _goalStructureStack.Pop();
             }
         }
+
+        /// <summary>
+        /// Implicitly lifts a goal into a desire set.
+        /// </summary>
+        /// <inheritdoc cref="LiftingExtensionMethods.Lift{TBeliefSet}(IGoal{TBeliefSet})" path="/param[@name='goal']"/>
+        /// <returns>The most logically matching desire set, wrapping around <paramref name="goal"/>.</returns>
+        public static implicit operator DesireSet<TBeliefSet>(Goal<TBeliefSet> goal) => goal.Lift();
```
JensSteenmetz: This is now implicitly converted, since goal.Lift() returns a goal structure. Could you do this explicitly for clarity?
```suggestion
        public static implicit operator DesireSet<TBeliefSet>(Goal<TBeliefSet> goal) => goal.Lift().Lift();
```

Aplib.Core.Tests/LiftingTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// Then the two should be equal, but the metadata name should differ.
+    /// </summary>
+    [Theory]
+    [MemberData(nameof(MetadataMemberData))]
+    public void ImplicitLifting_GoalStructureLiftedToDesireSet_DoesNotDifferFromManualTactic(Metadata? metadata)
+    {
+        // Arrange
+        GoalStructure<IBeliefSet> goalStructure = metadata is null
+            ? new PrimitiveGoalStructure<IBeliefSet>(It.IsAny<Goal<IBeliefSet>>())
+            : new PrimitiveGoalStructure<IBeliefSet>(metadata, It.IsAny<Goal<IBeliefSet>>());
```
JensSteenmetz: Conditionals in a test is a test smell, could you separate them into multiple tests? I suggest looking at [how Luka did it](https://github.com/team-zomsa/aplib.net/blob/dev/Aplib.Core.Tests/CombinatorTests.cs).

JensSteenmetz (changes requested):
Awesome work! And thanks for the Linq refactor :) I left a few comments.

Aplib.Core/LiftingExtensionMethods.cs
SilasPeters: I always include the metadata, for its a feature and something someone should rely on: if something's lifted, you should know. To rely on that, I do not make it optional.

Besides, if I implicitly lift, I would still need to check if you properly implemented the metadata. It partially moves the problem.

If you look at lifting manually, I am not in favor of optionally leaving out the previous metadata. It resets all info previously known, kind of destructive. Want another metadata? The user can set it himself. Does not matter if he does that as an argument, or by setting it manually aftewards, same number of characters (roughly).

Aplib.Core/Desire/DesireSets/DesireSet.cs
SilasPeters: So I should not use the feature I am implementing?

Aplib.Core.Tests/LiftingTests.cs
SilasPeters: Test smell, whilst it heavily reduces code duplication and no logic relies on this logic.

Aplib.Core/LiftingExtensionMethods.cs
JensSteenmetz: 1. That 'destructiveness' of the metadata is the same behavior that you get when not using lifting. For example, let's say you have an action A, and want to create a primitive tactic from it. The code `new PrimitiveTactic(A)` will create a primitive tactic without metadata. 
2. It also does not actually reset/destruct any data when lifting, because the underlying action still has its metadata after lifting with the lift method. No information is lost.
3. Cannot provide your own metadata when lifting. Sure, you can do it afterward, but why not give it as a parameter when you're at it? If we want to let the user do it afterward, then why not do so for _all_ other constructors as well?
4. If something's lifted, you _do_ know, because you lifted it.

Aplib.Core/Desire/DesireSets/DesireSet.cs
JensSteenmetz: My point is not about using features or not. In my opinion, calling Lift two times makes it immediately clear what the code is doing, while using one explicit lift and one implicit lift is not immediately clear I think. The first thing I checked when I looked at this code is if there was a Lift method lifting a Goal to a DesireSet (this is not even possible in C#, but I still looked for it). It just makes the code a little bit more clear IMO. Bonus: it removes a function call.

Aplib.Core.Tests/LiftingTests.cs
JensSteenmetz: @joachimdekker! I choose you!!!
pokemon-i-choose-you
@joachimdekker, use UNIT TEST KNOWLEDGE!!!!

Aplib.Core/IDocumented.cs
```diff
@@ -0,0 +1,16 @@
+namespace Aplib.Core
+{
+    /// <summary>
+    /// Represents an object that contains an instance of <see cref="IMetadata"/>.
+    /// </summary>
+    public interface IDocumented
+    {
+        /// <summary>
+        /// Gets the metadata of this BDI cycle component.
```
joachimdekker: Would `IDocumented` only be scoped at BDI cycle components. What if we add multiple agents, and we want to add documentation to those agents?

joachimdekker: The same goes for `IMetadata`: Try to provide descriptions to the interfaces that are as broad as possible. This way, the documentation could be inherited for any class that chooses to implement this interface.

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -19,40 +20,36 @@ public class FirstOfTactic<TBeliefSet> : AnyOfTactic<TBeliefSet>
...
-                IAction<TBeliefSet>? action = subTactic.GetAction(beliefSet);
-
-                if (action is not null) return action;
-            }
-
-            return null;
+            return _subTactics
+                .Select(subTactic => subTactic.GetAction(beliefSet))
+                .OfType<IAction<TBeliefSet>>()
+                .FirstOrDefault();
```
joachimdekker, line 52: EXCELLENT! How did we not think of this earlier. Incredible refactor!

Aplib.Core/LiftingExtensionMethods.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </returns>
+        /// <param name="goal">
+        /// The goal which on its own can function as a goal structure. Meaning, the goal structure consists of just a
+        /// single goal.
+        /// </param>
+        /// <param name="metadata">Optional metadata to be assigned to the goal structure.</param>
+        public static PrimitiveGoalStructure<TBeliefSet> Lift<TBeliefSet>(
+            this IGoal<TBeliefSet> goal,
+            IMetadata? metadata = null)
+            where TBeliefSet : IBeliefSet => metadata is null
```
joachimdekker: If I am correct, the metadata will never be null, as a default metadata will be set. Therefore, this ternary boolean expression is not needed, and metadata can just be passed.

joachimdekker (dismissed):
Excellent PR. I left one structural/architectural comment, the rest are nitpicks.

Aplib.Core.Tests/LiftingTests.cs
SilasPeters: @joachimdekker

JensSteenmetz (approved):
AWESOME

SSelinn (approved):
Nice

PR #53, opened 2024-05-13
feat!: restructure folders

joachimdekker:
## Description

This PR proposes to change the folder structure of aplib. It puts GoalStructure in its own folder, and creates a seperate namespace for agent.

BREAKING CHANGE: Because agent has a seperate namespace, this PR introduces a breaking change.

Closes AP-117.

## Checklist


Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -6,28 +6,15 @@ namespace Aplib.Core.Desire.Goals
 {
     /// <summary>
     /// A goal effectively combines a heuristic function with a tactic, and aims to meet the heuristic function by
-    /// applying the tactic. Goals are combined in a <see cref="GoalStructure{TBeliefSet}" />, and are used to prepare tests
-    /// or do
-    /// the testing.
+    /// applying the tactic. Goals are combined in a <see cref="GoalStructure.GoalStructure{TBeliefSet}" />, and are used to
+    /// prepare tests
+    /// or do the testing.
```
SilasPeters: grrrrr fix this :(((((((((((

SilasPeters (changes requested):

Aplib.Core/Desire/Goals/Goal.cs
joachimdekker: Rider won't let me :(

SilasPeters (dismissed):
You did not give any formal proofs. However, I wouldn't even check them if you did.

Aplib.Core/Desire/Goals/Goal.cs
QuakeEye: common Rider L

Aplib.Tests/Desire/GoalStructureTests.cs
```diff
@@ -1,9 +1,11 @@
 using Aplib.Core.Belief;
 using Aplib.Core.Desire;
 using Aplib.Core.Desire.Goals;
+using Aplib.Core.Desire.GoalStructure;
 using FluentAssertions;
 using Moq;
 using Moq.Protected;
+using System;
```
JensSteenmetz: Nit: could this using statement be removed? I would like to prevent using System namespace in aplib code, since the System.Action delegate clashes with Aplib.Core.Intent.Actions.Action. Even though Aplib.Core.Intent.Actions is not used in this script currently, it might be used later.

Aplib.Core/Desire/Goals/Goal.cs
JensSteenmetz, line 1: Could this folder/namespace be renamed to Goal? To be consistent with the GoalStructure namespace? Same comment on Aplib.Core.Intent.Actions and Aplib.Core.Intent.Tactics

JensSteenmetz (commented):
Could you also introduce DesireSet, Belief.Belief and Belief.BeliefSet namespaces?

JensSteenmetz (changes requested):
Left a Comment instead of Request changes by accident.

JensSteenmetz:
To formally prove the correctness of our new folder structure, we introduce a formalized set-theoretic approach. Let's define our folder structure as a set $F$ and introduce a function $\phi$ that maps elements of $F$ to their respective responsibilities within the system.

**Axiom 1: Modularity (M)**
A folder structure $F$ is said to be modular if for every element $f \in F$, there exists a unique responsibility $r$ such that $\phi(f) = r$. This implies a one-to-one correspondence between folders and responsibilities.

**Axiom 2: Scalability (S)**
A folder structure $F$ is scalable if for any non-empty subset $G \subseteq F$, there exists an element $g \notin F$ such that $F \cup \{g\}$ is also a folder structure satisfying Axiom 1.

**Theorem: The folder structure $F$ is effective if it satisfies both Modularity (M) and Scalability (S).**

**Proof:**

Let $F = \{ Core, Agent, Belief, BeliefSet, Desire, GoalStructure, Goal, DesireSet, Intent, Action, Tactic, Collections \}$.

**Step 1: Verify Modularity**
Define $\phi: F \rightarrow R$ where $R$ is the set of responsibilities. For instance, $\phi(Agent)$ could be 'Agent Operations', $\phi(Belief)$ could be 'Belief Management', and so on.

Since each folder in $F$ has a distinct responsibility, and no two folders share the same responsibility, $\phi$ is injective, thus satisfying Modularity (M).

**Step 2: Verify Scalability**
Consider a subset $G \subseteq F$, say $G = \{ Belief, BeliefSet \}$. If we want to add a new belief $b$, we can do so by creating a new element $g \notin F$, such that $g = b$ and $\phi(g) = 'New Belief'$.

Since adding $g$ to $F$ does not require changing the existing structure and $\phi$ remains injective, $F$ satisfies Scalability (S).

By demonstrating that $F$ satisfies both Modularity (M) and Scalability (S), we conclude that the folder structure is effective according to our theorem.

Aplib.Core/Desire/Goals/Goal.cs
joachimdekker: I don't agree. It houses multiple `Goals` just like the `System.Collections`. I will change the `GoalStructure` folder to `GoalStructures`.

Aplib.Tests/Desire/GoalStructureTests.cs
joachimdekker: This is the test script. Is that really necessary?

SilasPeters (changes requested):
The Aplib.Tests project should be renamed to Aplib.Core.Tests, for every project needs its own tests project

Aplib.Core/Desire/Goals/Goal.cs
JensSteenmetz: That's also fine

Aplib.Tests/Desire/GoalStructureTests.cs
JensSteenmetz: It is not giving issues right now, but it's sort of personal preference. I don't like having using directives that could potentially cause issues down the line. But not necessary, hence the nit. So you can leave it if you disagree.

Aplib.Core/Agents/BdiAgent.cs
```diff
@@ -4,14 +4,17 @@
...
 {
     /// <summary>
     /// Represents an agent that performs actions based on goals and beliefs.
     /// </summary>
     public class BdiAgent<TBeliefSet> : IAgent
         where TBeliefSet : IBeliefSet
     {
+        /// <inheritdoc />
+        public CompletionStatus Status => _desireSet.Status;
+
```
JensSteenmetz: Read-only should be above non-read-only public fields and properties

JensSteenmetz (changes requested):
One small thing regarding our conventions, lgtm otherwise

JensSteenmetz (approved):
LGTMMMM

SilasPeters (approved):

PR #54, opened 2024-05-13
fix!: improve constructors

Laura-Berkers:
## Description

This PR includes the following fixes:

- Add Metadata to some components that did not have it yet but should have.
- Remove the use of the nullable "hack" to make parameters optional, instead use overloads, nullable parameters are only used if they semantically make sense.
- Make the metadata truly optional in all constructors.
  - Make it possible to instantiate classes using `params` in the constructors without having to write `null` for the metadata.
  - Move the metadata to the front of all constructors.
- For all classes, put the construction logic in the constructor with the most amount of parameters, then define overloaded constructors that call the big constructor with default values to realize optional parameters. Some classes may have multiple constructor hierarchies of they have different kind of parameters.
- Use inheritdoc on constructors to prevent documentation duplication that may fall out of sync. 

## Breaking change

Because the parameters of constructors are moved around.

## Acceptance Criteria

-   The _guard field in Tactic should be able to be made read-only.

-   As an example, the following syntax should be changed from:

    ```cs
    public AnyOfTactic(System.Func<TBeliefSet, bool> guard, Metadata? metadata = null, params ITactic<TBeliefSet>[] subTactics)
      : this(metadata, subTactics) => _guard = guard;
    ```

    To multiple constructors like so:

    ```cs
    public AnyOfTactic(System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
      : this(null, guard, subTactics) {}

    public AnyOfTactic(Metadata? metadata, System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
      : base(metadata, guard) { // Set subtactics. }
    ```

-   It should be possible to call a tactic or goalstructure without specifying metadata (or null)

-   All Aplib components should be able to get metadata in this way

## Checklist
- [x] Merged dev into your branch


Aplib.Core/Belief/Belief.cs
```diff
@@ -30,57 +30,88 @@ public class Belief<TReference, TObservation> : IBelief where TReference : class
...
-        /// that this is not enforced by C# if <typeparamref name="TReference"/> is an interface.
-        /// </param>
-        /// <param name="getObservationFromReference">A function that takes an object reference and generates/updates an observation.</param>
-        /// <param name="shouldUpdate">A condition on when the observation should be updated.</param>
-        /// <exception cref="ArgumentException">
-        /// Thrown when <paramref name="reference"/> is not a reference type.
-        /// </exception>
-        public Belief(TReference reference,
+        /// <inheritdoc
+        ///     cref="Belief{TReference,TObservation}(Aplib.Core.Metadata,TReference,Func{TReference,TObservation},Func{bool})"/>
```
SilasPeters: This mentions metadata, but it is not part of the consturctor

Aplib.Core/Belief/Belief.cs
```diff
@@ -30,57 +30,88 @@ public class Belief<TReference, TObservation> : IBelief where TReference : class
...
-            Func<bool> shouldUpdate)
-            : this(reference, getObservationFromReference)
+            Func<bool> shouldUpdate
+        )
+            : this(new Metadata(), reference, getObservationFromReference, shouldUpdate)
+        {
+        }
+
+        /// <inheritdoc
+        ///     cref="Belief{TReference,TObservation}(Aplib.Core.Metadata,TReference,Func{TReference,TObservation},Func{bool})" />
```
SilasPeters: This mentions the predicate, which is not present.

Aplib.Core/Belief/ListBelief.cs
```diff
@@ -23,49 +23,62 @@ public class ListBelief<TReference, TObservation> : Belief<IEnumerable<TReferenc
...
-        /// <i>must</i> be a reference type, note that this is not enforced by C#.
-        /// </param>
-        /// <param name="getObservationFromReference">
-        /// A function that takes an object reference and generates an observation.
-        /// </param>
-        /// <param name="shouldUpdate">A condition on when the observation should be updated.</param>
-        /// <exception cref="ArgumentException">
-        /// Thrown when <paramref name="references"/> is not a reference type.
-        /// </exception>
+        /// <inheritdoc cref="ListBelief{TReference,TObservation}(Metadata,IEnumerable{TReference},Func{TReference,TObservation},Func{bool})"/>
```
SilasPeters: Again this mentions MetaData, and thus inherits and displays documentation for a param which is not present. I will not mention it any more, but I request you to remove these redundant xml docs everywhere.

Aplib.Core/Belief/SampledMemoryBelief.cs
```diff
@@ -46,59 +46,51 @@ private int ModuloCounter
...
         /// <param name="reference">
         /// The reference used to generate/update the observation. This <i>must</i> be a reference type.
         /// </param>
-        /// <param name="getObservationFromReference">A function that takes a reference and generates/updates a observation.</param>
+        /// <param name="getObservationFromReference">
+        /// A function that takes a reference and generates/updates a observation.
+        /// </param>
         /// <param name="sampleInterval">
-        /// The sample interval of the memory.
+        /// The sample rate of the memory.
```
SilasPeters: The rate (20 times per second) is not the same as interval (50ms)

Aplib.Core/Intent/Tactics/AnyOfTactic.cs
```diff
@@ -16,33 +17,44 @@ public class AnyOfTactic<TBeliefSet> : Tactic<TBeliefSet>
...
+        public AnyOfTactic
+        (
+            Metadata metadata,
+            Func<TBeliefSet, bool> guard,
+            params ITactic<TBeliefSet>[] subTactics
+        )
             : base(metadata)
         {
-            _subTactics = new();
+            _guard = guard;
```
SilasPeters: Shouldn't this be passed down to base?

SilasPeters (changes requested):
Damn, what a tedious PR. God bless you.

Aplib.Core/Belief/Belief.cs
joachimdekker: That is the magic of `inheritdoc`

joachimdekker: This means we don't have to write every param again, removing docs duplication.

joachimdekker (dismissed):
What a PR! Looks all good!

Aplib.Core/Belief/Belief.cs
JensSteenmetz: Rider shows it anyway if I'm not mistaken, but that's a bug in Rider.

Aplib.Tests/Belief/MemoryBeliefTests.cs
```diff
@@ -8,6 +9,26 @@ namespace Aplib.Core.Tests.Belief;
 /// </summary>
 public class MemoryBeliefTests
 {
+    /// <summary>
+    /// Given no metadata,
+    /// When a MemoryBelief is constructed,
+    /// Then it gets a random id and no name or description.
+    /// </summary>
+    [Fact]
+    public void MemoryBelief_ConstructedWithoutMetadat_ContainsDefaultMetadata()
```
JensSteenmetz: Nit
```suggestion
    public void MemoryBelief_ConstructedWithoutMetadata_ContainsDefaultMetadata()
```

Aplib.Core/Intent/Tactics/PrimitiveTactic.cs
```diff
@@ -16,41 +16,55 @@ public class PrimitiveTactic<TBeliefSet> : Tactic<TBeliefSet>
...
         /// <param name="metadata">
         /// Metadata about this tactic, used to quickly display the tactic in several contexts.
         /// </param>
-        public PrimitiveTactic(IQueryable<TBeliefSet> queryAction,
-            System.Func<TBeliefSet, bool> guard,
-            Metadata? metadata = null)
-            : base(beliefSet => guard(beliefSet) && queryAction.Query(beliefSet), metadata)
-        => _action = queryAction;
+        /// <param name="queryAction">The queryable action of the primitive tactic.</param>
+        /// <param name="guard">The guard of the primitive tactic.</param>
```
JensSteenmetz: You can use inheritdoc and just add the line
`/// <param name="queryAction">The queryable action of the primitive tactic.</param>`

JensSteenmetz (changes requested):
Awesome work! I left two small comments.

Aplib.Core/Belief/ListBelief.cs
Laura-Berkers: This is a Rider bug. It renders correctly in docfx and in Visual Studio. I won't do a workaround for this Rider bug because the only workaround would involve an enormous amount of documentation duplication.

Aplib.Core/Belief/Belief.cs
SilasPeters: That's indeed why I mentioned it, Rider. And then again, not so much magic if you have to exhaustively mention everything. What if a field is added? Then, we need to manually expand the inherit doc. Isn't it possible to just inherit all?

Aplib.Core/Intent/Tactics/PrimitiveTactic.cs
Laura-Berkers: I'm not sure if that is a good idea because it takes the Rider bug from being potentially confusing to definitely very confusing. And the parameter ordering is then wrong.

JensSteenmetz: You have a point, but I'm not sure if we should work around Rider bugs. Inheritdoc is meant to support things like this. The parameter ordering is also a Rider thing I think. But we can leave it like this.

Aplib.Core/Belief/Belief.cs
JensSteenmetz: Right now it is inheriting all, if I understand you correctly. Inheritdoc takes all applicable XML tags on the referenced member. We don't do this to prevent work for ourselves in the short run, but to prevent missed docs changes if we ever want to change the XML docs on a member. With inheritdoc, we only need to change the docs in one place, and all other overloads would then also use the updated XML docs.

Regarding the addition of a field: see the discussion below. Link: https://github.com/team-zomsa/aplib.net/pull/54#discussion_r1606085548

Aplib.Core/Intent/Tactics/PrimitiveTactic.cs
JensSteenmetz: Could you test the parameter ordering in docfx?

JensSteenmetz (dismissed):
Awesome!

JensSteenmetz (dismissed):
LGTM!

Aplib.Core/Belief/Belief.cs
SilasPeters: Having to say `inheritdoc cref="fieldA, fieldB, fieldC" is not the same as including 'all', just explicitly every current field. If field D is added someday, it will not be included. I will resolve this to move on

Aplib.Core/Belief/Beliefs/SampledMemoryBelief.cs
```diff
@@ -46,70 +46,109 @@ private int ModuloCounter
...
-        public SampledMemoryBelief(
+        /// <inheritdoc
+        ///     cref="SampledMemoryBelief{TReference,TObservation}(Metadata,TReference,Func{TReference,TObservation},int,UpdateMode,int,Func{bool})" />
+        public SampledMemoryBelief
+        (
             TReference reference,
             Func<TReference, TObservation> getObservationFromReference,
             int sampleInterval,
             UpdateMode updateMode,
             int framesToRemember,
```
SilasPeters, line 100: How come this gets documentation?

SilasPeters (changes requested):
Tip: use `git log --remerge-diff` to see the changes Luka made after applying the initial merge - which caused merge conflicts. In other words, this shows how the merge conflict was solved in a condensed manner.

Aplib.Core/Belief/BeliefSet/BeliefSet.cs
JensSteenmetz, line 1: The folder name should be BeliefSets.

Aplib.Core/Belief/Beliefs/ListBelief.cs
```diff
@@ -23,47 +23,62 @@ public class ListBelief<TReference, TObservation> : Belief<IEnumerable<TReferenc
...
+        )
+            : this(metadata, references, getObservationFromReference, () => true)
+        {
+        }
+
+        /// <inheritdoc
+        ///     cref="ListBelief{TReference,TObservation}(Metadata,IEnumerable{TReference},Func{TReference,TObservation},Func{bool})" />
+        public ListBelief
+            (IEnumerable<TReference> references, Func<TReference, TObservation> getObservationFromReference)
+            : this(new Metadata(), references, getObservationFromReference)
```
JensSteenmetz:
```suggestion
            : this(new Metadata(), references, getObservationFromReference, () => true)
```

Aplib.Core/Belief/Beliefs/Belief.cs
```diff
@@ -30,57 +30,88 @@ public class Belief<TReference, TObservation> : IBelief where TReference : class
...
+            Func<TReference, TObservation> getObservationFromReference
+        )
+            : this(metadata, reference, getObservationFromReference, () => true)
+        {
+        }
+
+        /// <inheritdoc
+        ///     cref="Belief{TReference,TObservation}(Aplib.Core.Metadata,TReference,Func{TReference,TObservation},Func{bool})" />
+        public Belief(TReference reference, Func<TReference, TObservation> getObservationFromReference)
+            : this(new Metadata(), reference, getObservationFromReference)
```
JensSteenmetz: I would try to prevent too much constructor chaining. It results in a little bit of code duplication, but if we don't do it, then
- The reader would need to move all around the file to find all the constructors that are being called. 
- There would be more function calls (not really a reason to prevent it per se, but a small extra con).
- The following scenario would not have a clear implementation:
```cs
public Foo(Bar a, Baz b, Qux c) { // Do stuff with a, b, c. }
public Foo(Bar a, Baz b) : this(a, b, null) { }
public Foo(Bar a, Qux c) : this(a, null, c) { }
public Foo(Bar a) : this(a, null) { } // Ambiguous. Also, if we would not give null, but an actual unambiguous value for the second parameter, then we would be able to choose whether we give a Baz or a Qux; there would be no reason to choose one over the other.
```
```suggestion
            : this(new Metadata(), reference, getObservationFromReference, () => true)
```

Aplib.Core/Belief/Beliefs/MemoryBelief.cs
```diff
@@ -26,54 +26,69 @@ public class MemoryBelief<TReference, TObservation> : Belief<TReference, TObserv
...
+        )
+            : this(metadata, reference, getObservationFromReference, framesToRemember, () => true)
+        {
+        }
+
+        /// <inheritdoc
+        ///     cref="MemoryBelief{TReference,TObservation}(Metadata,TReference,Func{TReference,TObservation},int,Func{bool})" />
+        public MemoryBelief
+            (TReference reference, Func<TReference, TObservation> getObservationFromReference, int framesToRemember)
+            : this(new Metadata(), reference, getObservationFromReference, framesToRemember)
```
JensSteenmetz:
```suggestion
            : this(new Metadata(), reference, getObservationFromReference, framesToRemember, () => true)
```

Aplib.Core/Belief/Beliefs/SampledMemoryBelief.cs
```diff
@@ -46,70 +46,109 @@ private int ModuloCounter
...
         /// </summary>
+        /// <param name="metadata">
+        /// Metadata about this goal, used to quickly display the goal in several contexts.
+        /// </param>
         /// <param name="reference">
         /// The reference used to generate/update the observation. This <i>must</i> be a reference type.
         /// </param>
-        /// <param name="getObservationFromReference">A function that takes a reference and generates/updates a observation.</param>
+        /// <param name="getObservationFromReference">
+        /// A function that takes a reference and generates/updates a observation.
```
JensSteenmetz: Nit (also, not your fault)
```suggestion
        /// A function that takes a reference and generates/updates an observation.
```

Aplib.Core/Belief/Beliefs/SampledMemoryBelief.cs
```diff
@@ -46,70 +46,109 @@ private int ModuloCounter
...
+        ///     cref="SampledMemoryBelief{TReference,TObservation}(Metadata,TReference,Func{TReference,TObservation},int,UpdateMode,int,Func{bool})" />
+        public SampledMemoryBelief
+        (
+            TReference reference,
+            Func<TReference, TObservation> getObservationFromReference,
+            int sampleInterval,
+            UpdateMode updateMode,
+            int framesToRemember
+        )
+            : this(new Metadata(), reference, getObservationFromReference, sampleInterval, updateMode, framesToRemember)
```
JensSteenmetz:
```suggestion
            : this(new Metadata(), reference, getObservationFromReference, sampleInterval, updateMode, framesToRemember, () => true)
```

Aplib.Core/Desire/GoalStructures/RepeatGoalStructure.cs
```diff
@@ -18,12 +18,16 @@ public class RepeatGoalStructure<TBeliefSet> : GoalStructure<TBeliefSet>
...
-        /// <param name="goalStructure">The goalstructure to repeat</param>
-        public RepeatGoalStructure(IGoalStructure<TBeliefSet> goalStructure)
-            : base(
-                new List<IGoalStructure<TBeliefSet>> { goalStructure }
-            ) =>
-            _currentGoalStructure = goalStructure;
+        /// <param name="metadata">
+        /// Metadata about this goal, used to quickly display the goal in several contexts.
+        /// </param>
+        /// <param name="goalStructure">The GoalStructure to repeat</param>
```
JensSteenmetz: GOTCHA!
```suggestion
        /// <param name="goalStructure">The GoalStructure to repeat.</param>
```

Aplib.Core/Desire/GoalStructures/SequentialGoalStructure.cs
```diff
@@ -24,20 +24,25 @@ public class SequentialGoalStructure<TBeliefSet> : GoalStructure<TBeliefSet>, ID
...
-            : base(children)
+        public SequentialGoalStructure(Metadata metadata, params IGoalStructure<TBeliefSet>[] children)
+            : base(metadata, children)
         {
             if (children.Length <= 0)
-            {
                 throw new ArgumentException("Collection of children is empty", nameof(children));
-            }
-
             _childrenEnumerator = _children.GetEnumerator();
```
JensSteenmetz, line 37: Nit: keep the newline?
```suggestion
            
            _childrenEnumerator = _children.GetEnumerator();
```

Aplib.Core/Desire/GoalStructures/SequentialGoalStructure.cs
```diff
@@ -49,11 +54,7 @@ public override void UpdateStatus(TBeliefSet beliefSet)
             // This loop is here to prevent tail recursion.
             while (true)
             {
-                if (Status == CompletionStatus.Success)
-                {
-                    return;
-                }
-
+                if (Status == CompletionStatus.Success) return;
```
JensSteenmetz, line 58: Nit: keep the newline?
```suggestion
                if (Status == CompletionStatus.Success) return;
                
```

Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -41,65 +59,68 @@ public class Goal<TBeliefSet> : IGoal<TBeliefSet>
...
-        protected readonly HeuristicFunction _heuristicFunction;
-
         /// <summary>
         /// Creates a new goal which works with <see cref="Heuristics" />.
         /// </summary>
+        /// <param name="metadata">
+        /// Metadata about this goal, used to quickly display the goal in several contexts.
+        /// </param>
         /// <param name="tactic">The tactic used to approach this goal.</param>
         /// <param name="heuristicFunction">The heuristic function which defines whether a goal is reached</param>
```
JensSteenmetz: Nit
```suggestion
        /// <param name="heuristicFunction">The heuristic function which defines whether a goal is reached.</param>
```

Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -41,65 +59,68 @@ public class Goal<TBeliefSet> : IGoal<TBeliefSet>
...
         /// </param>
-        public Goal(
+        public Goal
+        (
+            Metadata metadata,
             ITactic<TBeliefSet> tactic,
             HeuristicFunction heuristicFunction,
-            double epsilon = 0.005d,
-            Metadata? metadata = null
+            double epsilon = _defaultEpsilon
```
JensSteenmetz: Why do we need to store this in a variable? I'd argue it does not make it less of a 'magic number', but decreases readability because the reader cannot immediately see the value in his IDE now.

JensSteenmetz: Okay, I understand why you do it actually. I still don't think it's worth it, because of readability.

Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -41,65 +59,68 @@ public class Goal<TBeliefSet> : IGoal<TBeliefSet>
...
-        /// The goal is considered to be completed, when the distance of the <see cref="DetermineCurrentHeuristics" /> is below
-        /// this value.
-        /// </param>
         /// <param name="metadata">
         /// Metadata about this goal, used to quickly display the goal in several contexts.
         /// </param>
-        public Goal(ITactic<TBeliefSet> tactic,
+        /// <param name="tactic">The tactic used to approach this goal.</param>
+        /// <param name="predicate">
+        /// The heuristic function (or specifically predicate) which defines whether a goal is reached
```
JensSteenmetz:
```suggestion
        /// The heuristic function (or specifically predicate) which defines whether a goal is reached.
```

Aplib.Core/Intent/Actions/Action.cs
```diff
@@ -25,28 +25,24 @@ public class Action<TBeliefSet> : IAction<TBeliefSet>
...
-            Metadata = metadata ?? new Metadata();
-        }
+        /// <inheritdoc cref="Action{TBeliefSet}(Aplib.Core.Metadata,System.Action{TBeliefSet})" />
+        public Action(System.Action<TBeliefSet> effect) : this(new Metadata(), effect) { }
+
+        /// <inheritdoc cref="Action{TBeliefSet}(Aplib.Core.Metadata,System.Action{TBeliefSet})" />
+        protected Action(Metadata metadata) : this(metadata, _ => { }) { }
+
+        /// <inheritdoc cref="Action{TBeliefSet}(Aplib.Core.Metadata,System.Action{TBeliefSet})" />
+        protected Action() : this(new Metadata()) { }
```
JensSteenmetz: We can completely remove these. I don't know why we did this and if it was a problem earlier, but we only added these to allow `QueryAction(...) : base(metadata)`. However, we can do this like so: `QueryAction(...) : base(metadata, _ => {})`
```suggestion
```

Aplib.Core/Intent/Actions/QueryAction.cs
```diff
@@ -29,20 +29,26 @@ public class QueryAction<TBeliefSet, TQuery> : Action<TBeliefSet>, IQueryable<TB
...
         /// Metadata about this action, used to quickly display the action in several contexts.
         /// </param>
-        public QueryAction(System.Action<TBeliefSet, TQuery> effect,
-            System.Func<TBeliefSet, TQuery?> query,
-            Metadata? metadata = null)
+        /// <param name="effect">The effect of the action.</param>
+        /// <param name="query">The query of the action.</param>
+        public QueryAction
+            (Metadata metadata, System.Action<TBeliefSet, TQuery> effect, System.Func<TBeliefSet, TQuery?> query)
             : base(metadata)
```
JensSteenmetz: Ties into previous comment.
```suggestion
            : base(metadata, _ => {})
```

Aplib.Core/Intent/Tactics/AnyOfTactic.cs
```diff
@@ -16,35 +16,40 @@ public class AnyOfTactic<TBeliefSet> : Tactic<TBeliefSet>
...
-        /// Metadata about this tactic, used to quickly display the tactic in several contexts.
-        /// </param>
-        /// <param name="subTactics">The list of sub-tactics.</param>
-        public AnyOfTactic(System.Func<TBeliefSet, bool> guard,
-            Metadata? metadata = null,
-            params ITactic<TBeliefSet>[] subTactics)
-            : this(metadata, subTactics) => _guard = guard;
+        /// <inheritdoc cref="AnyOfTactic{TBeliefSet}(Metadata,System.Func{TBeliefSet,bool},ITactic{TBeliefSet}[])" />
+        public AnyOfTactic(params ITactic<TBeliefSet>[] subTactics)
+            : this(new Metadata(), subTactics)
```
JensSteenmetz:
```suggestion
            : this(new Metadata(), _ => true, subTactics)
```

Aplib.Core/Intent/Tactics/AnyOfTactic.cs
```diff
@@ -16,35 +16,40 @@ public class AnyOfTactic<TBeliefSet> : Tactic<TBeliefSet>
...
         /// <summary>
-        /// Initializes a new instance of the <see cref="AnyOfTactic{TBeliefSet}"/> class with the specified sub-tactics.
+        /// Initializes a new instance of the <see cref="AnyOfTactic{TBeliefSet}"/> class with the specified sub-tactics
+        /// and an optional guard condition.
         /// </summary>
         /// <param name="metadata">
         /// Metadata about this tactic, used to quickly display the tactic in several contexts.
         /// </param>
+        /// <param name="guard">The guard condition.</param>
         /// <param name="subTactics">The list of sub-tactics.</param>
```
JensSteenmetz: Not-your-fault-nit
```suggestion
        /// <param name="subTactics">The list of subtactics.</param>
```

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -10,32 +10,36 @@ public class FirstOfTactic<TBeliefSet> : AnyOfTactic<TBeliefSet>
...
         /// <summary>
-        /// Initializes a new instance of the <see cref="FirstOfTactic{TBeliefSet}"/> class with the specified sub-tactics.
+        /// Initializes a new instance of the <see cref="FirstOfTactic{TBeliefSet}"/> class with the specified
+        /// sub-tactics and guard condition.
         /// </summary>
         /// <param name="metadata">
         /// Metadata about this tactic, used to quickly display the tactic in several contexts.
         /// </param>
+        /// <param name="guard">The guard condition.</param>
         /// <param name="subTactics">The list of sub-tactics.</param>
```
JensSteenmetz: Not-your-fault-nit
```suggestion
        /// <param name="subTactics">The list of subtactics.</param>
```

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -10,32 +10,36 @@ public class FirstOfTactic<TBeliefSet> : AnyOfTactic<TBeliefSet>
...
+
+        /// <inheritdoc cref="FirstOfTactic{TBeliefSet}(Metadata,System.Func{TBeliefSet,bool},ITactic{TBeliefSet}[])"/>
+        public FirstOfTactic
+            (System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
+            : this(new Metadata(), guard, subTactics)
         {
         }
 
+        /// <inheritdoc cref="FirstOfTactic{TBeliefSet}(Metadata,System.Func{TBeliefSet,bool},ITactic{TBeliefSet}[])"/>
+        public FirstOfTactic(params ITactic<TBeliefSet>[] subTactics) : this(_ => true, subTactics) { }
```
JensSteenmetz:
```suggestion
        public FirstOfTactic(params ITactic<TBeliefSet>[] subTactics) : this(new Metadata(), _ => true, subTactics) { }
```

Aplib.Core/Intent/Tactics/PrimitiveTactic.cs
```diff
@@ -16,41 +16,55 @@ public class PrimitiveTactic<TBeliefSet> : Tactic<TBeliefSet>
...
+        public PrimitiveTactic(IAction<TBeliefSet> action, System.Func<TBeliefSet, bool> guard)
+            : this(new Metadata(), action, guard)
+        {
+        }
+
+        /// <inheritdoc cref="PrimitiveTactic{TBeliefSet}(Metadata,IAction{TBeliefSet},System.Func{TBeliefSet,bool})"/>
+        public PrimitiveTactic(Metadata metadata, IAction<TBeliefSet> action) : this(metadata, action, _ => true) { }
+
+        /// <inheritdoc cref="PrimitiveTactic{TBeliefSet}(Metadata,IAction{TBeliefSet},System.Func{TBeliefSet,bool})"/>
+        public PrimitiveTactic(IAction<TBeliefSet> action) : this(new Metadata(), action) { }
```
JensSteenmetz:
```suggestion
        public PrimitiveTactic(IAction<TBeliefSet> action) : this(new Metadata(), action, _ => true) { }
```

Aplib.Core/Intent/Tactics/PrimitiveTactic.cs
```diff
@@ -16,41 +16,55 @@ public class PrimitiveTactic<TBeliefSet> : Tactic<TBeliefSet>
...
         /// Metadata about this tactic, used to quickly display the tactic in several contexts.
         /// </param>
-        public PrimitiveTactic(IQueryable<TBeliefSet> queryAction,
-            System.Func<TBeliefSet, bool> guard,
-            Metadata? metadata = null)
-            : base(beliefSet => guard(beliefSet) && queryAction.Query(beliefSet), metadata)
-            => _action = queryAction;
+        /// <param name="queryAction">The queryable action of the primitive tactic.</param>
+        /// <param name="guard">The guard of the primitive tactic.</param>
+        public PrimitiveTactic
```
JensSteenmetz, line 51: Could you also add constructors with IQueryable but guard always true?

Aplib.Core/Intent/Tactics/Tactic.cs
```diff
@@ -27,27 +27,28 @@ public abstract class Tactic<TBeliefSet> : ITactic<TBeliefSet>
...
         }
 
+        /// <inheritdoc cref="Tactic{TBeliefSet}(Aplib.Core.Metadata,System.Func{TBeliefSet,bool})" />
+        protected Tactic(System.Func<TBeliefSet, bool> guard) : this(new Metadata(), guard) { }
+
+        /// <inheritdoc cref="Tactic{TBeliefSet}(Aplib.Core.Metadata,System.Func{TBeliefSet,bool})" />
+        protected Tactic(Metadata metadata) : this(metadata, _ => true) { }
+
+        /// <inheritdoc cref="Tactic{TBeliefSet}(Aplib.Core.Metadata,System.Func{TBeliefSet,bool})" />
+        protected Tactic() : this(new Metadata()) { }
```
JensSteenmetz:
```suggestion
        protected Tactic() : this(new Metadata(), _ => true) { }
```

Aplib.Core/Metadata.cs
```diff
@@ -27,12 +27,7 @@ public class Metadata
         /// </summary>
         /// <param name="name">The name used to display the component.</param>
         /// <param name="description">The description used to describe the component.</param>
-        public Metadata(string? name = null, string? description = null)
-        {
-            Id = Guid.NewGuid();
-            Name = name;
-            Description = description;
-        }
+        public Metadata(string? name = null, string? description = null) : this(Guid.NewGuid(), name, description) { }
```
JensSteenmetz, line 30: Is there a specific reason not to do this? Then you don't need the internal constructor. I cannot give a full suggestion for the entire code that would be affected by this, unfortunately, but you get the idea.
```suggestion
        public Metadata(string? name = null, string? description = null)
        {
            Id = Guid.NewGuid();
            Name = name;
            Description = description;
        }
```

JensSteenmetz (changes requested):
I've left a few comments, but really great work on this PR!

Aplib.Core/Belief/Beliefs/SampledMemoryBelief.cs
JensSteenmetz: Resolved by @SilasPeters

SilasPeters (dismissed):

Aplib.Core/Desire/Goals/Goal.cs
Laura-Berkers: It is used in four places currently. I think the greater risk is that if we ever want to change the default, we forget to update it in some place.

SilasPeters: I agree with Jens and Luka

Aplib.Core/Metadata.cs
Laura-Berkers: The internal constructor was already there. We wanted to be able to give a specific id to some component.

Aplib.Core/Desire/Goals/Goal.cs
JensSteenmetz: We appreciate that

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Aplib.Core.Belief.BeliefSets;
 using Aplib.Core.Intent.Actions;
+using System;
```
JensSteenmetz: System.Action clashes with Aplib.Core.Intent.Actions.Action, so I would like to prevent using System in Aplib.Core at all.
```suggestion
```

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -17,9 +18,9 @@ public class FirstOfTactic<TBeliefSet> : AnyOfTactic<TBeliefSet>
         /// Metadata about this tactic, used to quickly display the tactic in several contexts.
         /// </param>
         /// <param name="guard">The guard condition.</param>
-        /// <param name="subTactics">The list of sub-tactics.</param>
+        /// <param name="subTactics">The list of subtactics.</param>
         public FirstOfTactic
-            (Metadata metadata, System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
+            (Metadata metadata, Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
```
JensSteenmetz:
```suggestion
            (Metadata metadata, System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
```

Aplib.Core/Intent/Tactics/FirstOfTactic.cs
```diff
@@ -32,13 +33,13 @@ public FirstOfTactic(Metadata metadata, params ITactic<TBeliefSet>[] subTactics)
 
         /// <inheritdoc cref="FirstOfTactic{TBeliefSet}(Metadata,System.Func{TBeliefSet,bool},ITactic{TBeliefSet}[])"/>
         public FirstOfTactic
-            (System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
+            (Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
```
JensSteenmetz:
```suggestion
            (System.Func<TBeliefSet, bool> guard, params ITactic<TBeliefSet>[] subTactics)
```

JensSteenmetz (changes requested):
Honestly amazing job. Couldn't have done it better. I left one last comment

Aplib.Core.Tests/Collections/ExposedQueueTests.cs
```diff
@@ -33,7 +33,9 @@ public void ExposedQueue_WhenInitialized_ShouldHaveCorrectMaxCountAndCount()
     public void ExposedQueue_WhenInitializedWithArray_ShouldHaveCorrectMaxCountAndCount(int[] array, int? countParam, int expectedMaxCount, int expectedCount)
     {
         // Arrange
-        ExposedQueue<int> queue = new(array, countParam);
+        ExposedQueue<int> queue = countParam.HasValue
+            ? new ExposedQueue<int>(array, (int)countParam)
+            : new ExposedQueue<int>(array);
```
joachimdekker: Having conditionals in tests is considered a [test smell](https://testsmells.org/pages/testsmells.html#ConditionalTestLogic). 

You should provide different tests with or without the countparam (this is as easy as just copying the test, changing the name and the params)

Aplib.Core/Collections/ExposedQueue.cs
```diff
@@ -65,7 +65,16 @@ public ExposedQueue(T[] array, int? count = null)
...
+        }
+
+        /// <inheritdoc cref="ExposedQueue{T}(T[],int)"/>
+        public ExposedQueue(T[] array)
+        {
+            MaxCount = array.Length;
+            _array = array;
+            _head = MaxCount - 1;
+            Count = MaxCount;
         }
```
joachimdekker: To stay up-to date with the other constructor, it would be better to call `this` from this constructor.

```suggestion
        /// <inheritdoc cref="ExposedQueue{T}(T[],int)"/>
        public ExposedQueue(T[] array)
            : this(array, array.Length)
        {
        }
```

Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -16,6 +15,25 @@ namespace Aplib.Core.Desire.Goals
     public class Goal<TBeliefSet> : IGoal<TBeliefSet>
         where TBeliefSet : IBeliefSet
     {
+        /// <summary>
+        /// The default value for the epsilon parameter in the Goal constructors.
+        /// The epsilon parameter defines the threshold distance for a goal to be considered completed.
+        /// </summary>
+        protected const double _defaultEpsilon = 0.005d;
```
joachimdekker: Constants should be named using PascalCase according to the notion wiki and the [CSharp naming conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions).

```suggestion
        protected const double DefaultEpsilon = 0.005d;
```
⚠️ This suggestion will break the build.

joachimdekker (changes requested):
I feel like shit commenting on this PR, as it is LEGENDARY... But I still found some. I tried to make them all suggestions 😄

Aplib.Core.Tests/Intent/Tactics/TacticTests.cs
```diff
@@ -49,6 +169,74 @@ public void AnyOfTacticGetAction_WhenASubtacticIsEnabled_ReturnsActionOfEnabledS
...
+        ITactic<IBeliefSet>[] subTactics = [It.IsAny<ITactic<IBeliefSet>>()];
+
+        // Act
+        TestFirstOfTactic tactic = new(subTactics);
+
+        // Assert
+        tactic.Metadata.Id.Should().NotBeEmpty();
+        tactic.Metadata.Name.Should().BeNull();
+        tactic.Metadata.Description.Should().BeNull();
+        tactic.Guard(It.IsAny<IBeliefSet>()).Should().BeTrue();
```
SilasPeters, line 236: This check is not done everywhere where guard is not null. Why is that?

Aplib.Core.Tests/Belief/BeliefTests.cs
```diff
@@ -30,6 +33,117 @@ public IEnumerator<int> GetEnumerator()
...
 
+    public class TestBelief : Belief<object, object>
+    {
+        public object Reference => _reference;
+
+        public Func<object, object> GetObservationFromReference => _getObservationFromReference;
+
+        public Func<bool> ShouldUpdate => _shouldUpdate;
+
+        public TestBelief
```
SilasPeters, line 44: If I understand correctly, you make these classes every time because you try to access protected variables.

If so, I suggest you add an internal method to Belief etc which exposes the protected variables for testing. Whilst protected fields cannot be seen by non-implementing types, internal variables can be exposed to referenced projects. Might reduce code coupling

joachimdekker: Wow, no, no, no. We don't add functionality just to make testing easier. That is such a bad practice.

joachimdekker: Code duplication or code coupling in tests is allowed, since they are just tests. In tests, it is better to define everything twice, since a test should be readable in a glance. The setup, act, and assert should be clear from one method. If we need a mock or belief to access internal variables, we do so via mocking, and not by editing the code under test.

Aplib.Core/Collections/ExposedQueue.cs
Laura-Berkers: Good call, I was looking at this yesterday but I guess my brain was too fried.

JensSteenmetz (approved):
<sup><sub>bruh...</sub></sup>

joachimdekker (approved):
thank_you_for_this_PR

QuakeEye:
> thank_you_for_this_PR

why an AI image?

PR #60, opened 2024-06-10
chore: add IDocumented to interfaces

JoenvandeVorle:
## Description

Adds IDocumented to interfaces for better accessibility 

## Testability

No new functional code

## Checklist

### DIFF
--- a/Aplib.Core.Tests/Belief/BeliefSetTests.cs
+++ b/Aplib.Core.Tests/Belief/BeliefSetTests.cs
@@ -134,6 +134,8 @@ public class BeliefSetTests
         /// </summary>
         public bool Updated { get; private set; } = false;
 
+        public IMetadata Metadata { get; } = new Metadata();
+
         /// <summary>
         /// Sets <see cref="Updated"/> to true.
         /// </summary>
--- a/Aplib.Core.Tests/Belief/BeliefSetTests.cs
+++ b/Aplib.Core.Tests/Belief/BeliefSetTests.cs
@@ -135,7 +135,15 @@ public class BeliefSetTests
         public bool Updated { get; private set; } = false;
 
         /// <inheritdoc />
-        public IMetadata Metadata { get; } = new Metadata("SimpleBelief");
+        public IMetadata Metadata { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimpleBelief"/> class.
+        /// </summary>
+        public SimpleBelief()
+        {
+            Metadata = new Metadata();
+        }
 
         /// <summary>
         /// Sets <see cref="Updated"/> to true.
--- a/Aplib.Core.Tests/Belief/BeliefSetTests.cs
+++ b/Aplib.Core.Tests/Belief/BeliefSetTests.cs
@@ -134,7 +134,8 @@ public class BeliefSetTests
         /// </summary>
         public bool Updated { get; private set; } = false;
 
-        public IMetadata Metadata { get; } = new Metadata();
+        /// <inheritdoc />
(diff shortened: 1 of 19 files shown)


LuukBerkers:
I think the constructors of Belief, ListBelief, MemoryBelief, and SampledMemoryBelief still need to be updated to take an IMetadata instead of a Metadata.

Laura-Berkers (commented):
I think the constructors of Belief, ListBelief, MemoryBelief, and SampledMemoryBelief still need to be updated to take an IMetadata instead of a Metadata.

Aplib.Core.Tests/Belief/BeliefSetTests.cs
```diff
@@ -134,6 +134,9 @@ private class SimpleBelief : IBelief
         /// </summary>
         public bool Updated { get; private set; } = false;
 
+        /// <inheritdoc />
+        public IMetadata Metadata { get; } = new Metadata("SimpleBelief");
```
SilasPeters: This does not match the current standards, and should be assigned in a constructor instead.

Laura-Berkers:
> I think the constructors of Belief, ListBelief, MemoryBelief, and SampledMemoryBelief still need to be updated to take an IMetadata instead of a Metadata.

Oh I see now this is an oversight in a previous PR, but could you do it here?

Aplib.Core/Belief/Beliefs/IBelief.cs
```diff
@@ -3,7 +3,7 @@
     /// <summary>
     /// A belief represents/encapsulates an observation (i.e., piece of information of the game state as perceived by an agent).
     /// </summary>
-    public interface IBelief
+    public interface IBelief : IDocumented
```
SilasPeters, line 6: I would advise against letting an interface implement another, unless both are concept-wise tightly bound together.

Otherwise, if someone were to implement his own `IBelief` for testing purposes, for example, he would be required to implement `IDocumented` as well.

In other words, it is not _always_ the case that, inheritly, a belief has documentation coming with it.

SilasPeters: For this reason, I let every implementation of `IBelief` inherit `IDocumented`.

If you somehow require that you can get an `IMetadata` from an `IBelief`, then I would suggest doing something like this:

```csharp
if (belief is IDocumented beliefDocumentation)
  var docs = beliefDocumentation;
```

or

```csharp
IDocumented? docs = belief as IDocumented; // Returns null if it does not implement IDocumented
```

SilasPeters (changes requested):

Aplib.Core.Tests/Belief/BeliefSetTests.cs
```diff
@@ -134,6 +134,17 @@ private class SimpleBelief : IBelief
         /// </summary>
         public bool Updated { get; private set; } = false;
```
Laura-Berkers, line 135:
```suggestion
        public bool Updated { get; private set; }
```

Aplib.Core.Tests/Belief/BeliefSetTests.cs
```diff
@@ -134,6 +134,17 @@ private class SimpleBelief : IBelief
...
 
+        /// <inheritdoc />
+        public IMetadata Metadata { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimpleBelief"/> class.
+        /// </summary>
+        public SimpleBelief()
+        {
+            Metadata = new Metadata();
```
Laura-Berkers, line 145:
```suggestion
            Updated = false;
            Metadata = new Metadata();
```

PR #61, opened 2024-06-10
fix: change behaviour of maxCount in ExposedQueue

joachimdekker:
## Description

This PR proposes to change the behavour of maxcount in the exposed queue constructor. 

Fixes AP-172.

## Testability

Run the tests :)

## Checklist

### DIFF
--- a/Aplib.Core/Collections/ExposedQueue.cs
+++ b/Aplib.Core/Collections/ExposedQueue.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Aplib.Core.Collections
 {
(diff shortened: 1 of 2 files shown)


Aplib.Core/Collections/ExposedQueue.cs
```diff
@@ -45,30 +46,27 @@ public ExposedQueue(int size)
...
-            if (count < 0)
-                throw new ArgumentOutOfRangeException(nameof(count), "Count cannot be negative.");
+            if (maxCount < 0) throw new ArgumentOutOfRangeException(nameof(maxCount), "Count cannot be negative.");
 
-            MaxCount = array.Length;
-            _array = array;
-            _head = MaxCount - 1;
-            Count = count;
+            MaxCount = maxCount;
+            _array = array.Take(maxCount).ToArray();
```
JensSteenmetz: `array.Take(maxCount).ToArray()` creates an array of `Min(maxCount, array.Length)` elements. However, we would like to have `maxCount` elements in the resulting array, where only `array.Length` slots/cells are filled.

JensSteenmetz (changes requested):

Aplib.Core/Collections/ExposedQueue.cs
joachimdekker: Excellent!

Aplib.Core/Collections/ExposedQueue.cs
```diff
@@ -45,30 +46,31 @@ public ExposedQueue(int size)
...
-            _array = array;
-            _head = MaxCount - 1;
-            Count = count;
+            MaxCount = maxCount;
+            _array = new T[maxCount];
+
+            if (array.Length > maxCount) array = array[..maxCount];
+
+            array.CopyTo(_array, 0);
+            Count = array.Length;
```
JensSteenmetz: If array.Length > maxCount, the array is copied twice. Can this be done more efficiently?
For example 
```suggestion
            Count = Math.Min(array.Length, MaxCount);
            
            _array = new T[maxCount];
            Array.Copy(array, _array, Count);
```

JensSteenmetz (changes requested):
sorry :)

JensSteenmetz (dismissed):
LGTM!

Aplib.Core/Collections/ExposedQueue.cs
```diff
@@ -45,30 +46,30 @@ public ExposedQueue(int size)
...
-        /// By default, assumes the array is filled. 
+        /// By default, assumes the array is filled.
         /// </summary>
         /// <param name="array">An array to use as the circular array.</param>
-        /// <param name="count">The number of actual elements in the array.</param>
+        /// <param name="maxCount">The number of actual elements in the array.</param>
         /// <remarks>
         /// The MaxCount of the queue will be set to the length of the array.
         /// If the array is not fully filled, the Count should be specified.
         /// </remarks>
```
JoenvandeVorle: Documentation is out of date now, should be updated

Aplib.Core/Collections/ExposedQueue.cs
```diff
@@ -45,30 +46,30 @@ public ExposedQueue(int size)
...
-            if (count < 0)
-                throw new ArgumentOutOfRangeException(nameof(count), "Count cannot be negative.");
+            if (maxCount < 0) throw new ArgumentOutOfRangeException(nameof(maxCount), "Count cannot be negative.");
 
-            MaxCount = array.Length;
-            _array = array;
-            _head = MaxCount - 1;
-            Count = count;
+            MaxCount = maxCount;
+            Count = Math.Min(array.Length, maxCount);
```
JoenvandeVorle, line 64: It seems now the constructor always assumes the array is full, there is no way to specify that you are passing an array where only x slots are filled.
How do we want this to handle empty arrays? (Open question, I don't know the best way)

JoenvandeVorle (changes requested):
left 2 comments :)

Aplib.Core/Collections/ExposedQueue.cs
JensSteenmetz: I'd say we don't handle them. If I'm passing an empty array, I'd expect the resulting ExposedQueue to have those empty spots as well. It doesn't break the code, so I'd leave it like this.

JoenvandeVorle: What happens is opposite though, it assumes the array is full even if you pass an empty one (the count just gets set to the size of the array). I talked it over with Joachim though and it seems okay to expect full arrays as long as that's clear from the summary.

JoenvandeVorle (approved):
lgtm

JensSteenmetz (approved):
LGTM

PR #62, opened 2024-06-10
test: add repeating test

joachimdekker:
## Description

This test adds a test repeating if a goal structure or goal can be repeated without having consequences.

## Testability

Check and run the added unit test

## Checklist
- [x] Merged dev into your branch

### DIFF
--- a/Aplib.Core.Tests/Desire/GoalStructureTests.cs
+++ b/Aplib.Core.Tests/Desire/GoalStructureTests.cs
@@ -10,21 +10,6 @@ namespace Aplib.Core.Tests.Desire;
 public class GoalStructureTests
 {
     [Fact]
-    public void PrimitiveGoalStructure_ConstructedWithMetadata_HasCorrectMetadata()
-    {
-        // Arrange
-        Metadata metadata = new("My GoalStructure", "The best GoalStructure");
-        Mock<IGoalStructure<IBeliefSet>> goalStructure1 = new();
-
-        // Act
-        FirstOfGoalStructure<IBeliefSet> goalStructure = new(metadata, goalStructure1.Object);
-
-        // Assert
-        goalStructure.Metadata.Name.Should().Be("My GoalStructure");
-        goalStructure.Metadata.Description.Should().Be("The best GoalStructure");
-    }
-
-    [Fact]
     public void FirstOfGoalStructure_WhenAllGoalsFail_ShouldReturnFailure()
     {
         Mock<IGoalStructure<IBeliefSet>> goalStructure1 = new();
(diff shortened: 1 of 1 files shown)


Aplib.Core.Tests/Desire/GoalStructureTests.cs
```diff
@@ -267,6 +267,36 @@ public void RepeatGoalStructure_WhenGoalStructureHasSucceeded_ShouldSucceed()
...
+
+        Mock<IGoalStructure<IBeliefSet>> intermediaryGoalStructureMock = new();
+        intermediaryGoalStructureMock.Setup(g => g.Status).Returns(CompletionStatus.Unfinished);
+
+        SequentialGoalStructure<IBeliefSet> sequentialGoalStructure =
+            new(primitiveGoalStructure, intermediaryGoalStructureMock.Object, primitiveGoalStructure);
+
+        // Act
+        sequentialGoalStructure.UpdateStatus(Mock.Of<IBeliefSet>());
+        intermediaryGoalStructureMock.Setup(g => g.Status).Returns(CompletionStatus.Success);
```
JensSteenmetz: Could you use SetupSequence, so that it can be put in // Arrange?

Aplib.Core.Tests/Desire/GoalStructureTests.cs
```diff
@@ -267,6 +267,36 @@ public void RepeatGoalStructure_WhenGoalStructureHasSucceeded_ShouldSucceed()
...
+        goal.Setup(g => g.GetStatus(It.IsAny<IBeliefSet>())).Returns(CompletionStatus.Unfinished);
```
JensSteenmetz: Same here

Aplib.Core.Tests/Desire/GoalStructureTests.cs
```diff
@@ -267,6 +267,36 @@ public void RepeatGoalStructure_WhenGoalStructureHasSucceeded_ShouldSucceed()
...
+        sequentialGoalStructure.UpdateStatus(Mock.Of<IBeliefSet>());
+
+        IGoal<IBeliefSet> currentGoal = sequentialGoalStructure.GetCurrentGoal(Mock.Of<IBeliefSet>());
+
+        // Assert
+        currentGoal.Should().Be(goal.Object);
+        intermediaryGoalStructureMock.Object.Status.Should().Be(CompletionStatus.Success);
+        primitiveGoalStructure.Status.Should().Be(CompletionStatus.Unfinished);
```
JensSteenmetz, line 333: By using a Mock, you're now not checking our Goal class. You defined the return of Status as CompletionStatus.Unfinished, so it's not testing the actual Goal behavior. The behavior we would like to have tested, is that the actual Goal class resets its status before being evaluated for the second time.

JensSteenmetz (changes requested):

Aplib.Core.Tests/Desire/GoalStructureTests.cs
joachimdekker: The test is not meant to check the Goal behaviour, but the behaviour of the `GoalStructure`. It is impossible to check the status of the goal class, as it will always be behind a `PrimitveGoalStructure`.

JensSteenmetz: Ahh okay. In that case I think we need to test it for the Goal class in a separate test, as now there's no guarantee that Goal has the behavior that the Moqs have here. So we should test in another test that Goal has that exact behavior of re-evaluating the status.

We could test this by calling GetStatus two times, where the first time the predicate returns Success, and the second time the predicate returns Failure.

JensSteenmetz: Nevermind, there is a test for this already!

JensSteenmetz (approved):
LGTM

SSelinn (approved):
Noice

PR #63, opened 2024-06-12
feat!: add fail-guard to Goal

Laura-Berkers:
## Description

This PR adds a failure condition to the Goal class. The used can provide a fail-guard upon construction of the Goal. If no fail-guard is provided, the goal will never fail.

The fail-guard is checked in de UpdateStatus method of the goal, before the heuristic. Thus, failure takes precedence over success.

This PR also changed all usages of `System.Func<T, bool>` to `System.Predicate<T>`, which is more semantically accurate.

And this PR also changes the `GetStatus` method in `IGoal` to `UpdateStatus` to be in line with other classes/interfaces implementing the `ICompletable` interface.

Lastly, this PR removes all remaining imports of the plain `System` namespace from the project, to avoid collision between `System.Action` and `Aplib.Core.Intent.Actions.Action`.

## Breaking change

- Because a `System.Func<T, bool>` cannot be implicitly converted to a `System.Predicate<T>`.
- Because the `GetStatus` method is replaced by `UpdateStatus`.

## Checklist

### DIFF
--- a/Aplib.Core.Tests/Desire/GoalTests.cs
+++ b/Aplib.Core.Tests/Desire/GoalTests.cs
@@ -172,4 +172,21 @@ public class GoalTests
         // Assert
         goalBooleanEvaluation.Should().Be(goalNonBooleanEvaluation);
     }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void UpdateStatus_WhenFailGuardIsTrue_FailsTheGoal(bool goalCompleted)
+    {
+        // Arrange
+        ITactic<IBeliefSet> tactic = Mock.Of<ITactic<IBeliefSet>>();
+        Goal<IBeliefSet>.HeuristicFunction heuristic = CommonHeuristicFunctions<IBeliefSet>.Boolean(_ => goalCompleted);
+        Goal<IBeliefSet> goal = new(tactic, heuristic, _ => true);
+
+        // Act
+        goal.UpdateStatus(It.IsAny<IBeliefSet>());
+
+        // Assert
+        goal.Status.Should().Be(CompletionStatus.Failure);
+    }
 }
--- a/Aplib.Core.Tests/Desire/GoalTests.cs
+++ b/Aplib.Core.Tests/Desire/GoalTests.cs
@@ -189,4 +189,21 @@ public class GoalTests
         // Assert
         goal.Status.Should().Be(CompletionStatus.Failure);
     }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void Goal_WithoutFailGuard_DoesNotFail(bool goalCompleted)
+    {
+        // Arrange
+        ITactic<IBeliefSet> tactic = Mock.Of<ITactic<IBeliefSet>>();
+        Goal<IBeliefSet>.HeuristicFunction heuristic = CommonHeuristicFunctions<IBeliefSet>.Boolean(_ => goalCompleted);
+
+        // Act
+        Goal<IBeliefSet> goal = new(tactic, heuristic);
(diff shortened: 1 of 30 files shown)


Aplib.Core/Belief/Beliefs/Belief.cs
```diff
@@ -1,6 +1,4 @@
-﻿using System;
```
JoenvandeVorle, line 1: So why did you remove this everywhere? I'm not too much of a fan of it

Laura-Berkers:
Updated PR description to mentions the removal of `System` imports.

Aplib.Core/Desire/Goals/Goal.cs
```diff
@@ -51,70 +56,143 @@ public class Goal<TBeliefSet> : IGoal<TBeliefSet>, IDocumented
         /// </summary>
         public ITactic<TBeliefSet> Tactic { get; }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Gets the completion status of the goal.
+        /// This value may need to be updated first using the <see cref="UpdateStatus"/> method.
+        /// </summary>
+        /// <seealso cref="UpdateStatus"/>
         public CompletionStatus Status { get; protected set; }
```
JoenvandeVorle, line 64: Maybe it's an idea to call UpdateStatus in the getter here before returning it?

Aplib.Core.Tests/Desire/GoalTests.cs
```diff
@@ -124,12 +126,15 @@ public void Goal_WhereHeuristicsChange_UsesUpdatedHeuristics()
         // Arrange
         IBeliefSet beliefSetMock = Mock.Of<IBeliefSet>();
         bool shouldSucceed = false;
+        // ReSharper disable once AccessToModifiedClosure
```
JoenvandeVorle: Reshaper leftover comment?

Aplib.Core.Tests/Desire/GoalTests.cs
```diff
@@ -150,18 +155,68 @@ public void GoalConstructor_WhereHeuristicFunctionTypeDiffers_HasEqualBehaviour(
         // Arrange
         ITactic<IBeliefSet> tactic = Mock.Of<ITactic<IBeliefSet>>();
 
-        bool heuristicFunctionBoolean(IBeliefSet _) => goalCompleted;
+        // ReSharper disable once MoveLocalFunctionAfterJumpStatement
```
JoenvandeVorle: same here

JoenvandeVorle (changes requested):

Aplib.Core/Belief/Beliefs/Belief.cs
Laura-Berkers: It causes a naming collision between `System.Action` and `Aplib.Core.Intent.Actions.Action`. This has already caused som confusion amongst ourselves when writing code, because you might think you are creating or using an aplib Action but you are actually using a System.Action. This can also created very confusing error messages.

Aplib.Core/Desire/Goals/Goal.cs
Laura-Berkers: Yes that's a good idea actually.

Laura-Berkers: I think actually we might remove the `UpdateStatus` method completely and just put it in the getter.

Laura-Berkers: And also from goal structures, for consistency.

JensSteenmetz (approved):
LGTM!!

JoenvandeVorle (approved):
lgtm

